PR #9, opened 2022-11-22
feat(lib): define the default install process for the plugin

vincenzopalazzo:
This allows storing information inside a plugin regarding the languages.

I'm still not sure how to get the default install information, maybe from a file or maybe just hardcoded. Not sure, but for now I think it is a good design of how to configure the process looks like!

Signed-off-by: Vincenzo Palazzo <vincenzopalazzodev@gmail.com>


swaptr (approved):
This looks good to me. I will add changes on top of this in the other PR.
Thank you. :heart:

PR #10, opened 2022-11-23
lib: implement URL for easy url handling

swaptr:
This PR adds support to abstract URL related functionalities to a separate struct.

### DIFF
--- a/reckless_github/src/lib.rs
+++ b/reckless_github/src/lib.rs
@@ -6,8 +6,9 @@ pub mod repository;
 mod tests {
     use std::{path::Path, sync::Once};
 
+    use log::debug;
     use reckless_lib::repository::Repository;
-    use reckless_lib::utils::get_dir_path_from_url;
+    use reckless_lib::url::URL;
 
     use crate::repository::Github;
 
@@ -24,10 +25,10 @@ mod tests {
     async fn repository_is_initialized_ok() {
         init();
         let name = "hello";
-        let url = "https://github.com/lightningd/plugins";
-        let mut repo = Github::new(name, url);
+        let url = URL::new("https://github.com/lightningd/plugins");
+        let mut repo = Github::new(name, &url);
         let repo = repo.init().await;
         assert!(repo.is_ok());
-        assert_eq!(Path::new(&(get_dir_path_from_url(url))).exists(), true);
+        assert_eq!(Path::new(&url.path_string).exists(), true);
     }
 }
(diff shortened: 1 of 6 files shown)


reckless_github/src/lib.rs
```diff
@@ -24,10 +25,13 @@ mod tests {
...
         let name = "hello";
-        let url = "https://github.com/lightningd/plugins";
-        let mut repo = Github::new(name, url);
+        let url = URL::new("https://github.com/lightningd/plugins");
+        let mut repo = Github::new(name, &url);
         let repo = repo.init().await;
         assert!(repo.is_ok());
-        assert_eq!(Path::new(&(get_dir_path_from_url(url))).exists(), true);
+        assert_eq!(
+            Path::new(&(get_dir_path_from_url(&url.url_string))).exists(),
```
vincenzopalazzo, line 33:
```suggestion
            Path::new(&(get_dir_path_from_url(&url.url_string))).exists(),
```
```suggestion
            Path::new(&url.path_string)).exists(),
```

I do not know if it is the same but we can implement the method  `get_dir_path_from_url` inside the URL?

vincenzopalazzo: maybe a good place to put a FIXME?

reckless_github/src/repository.rs
```diff
@@ -112,13 +113,16 @@ impl Repository for Github {
         debug!(
             "INITIALIZING REPOSITORY: {} {} > {}",
             self.name,
-            self.url,
-            get_dir_path_from_url(&self.url)
+            self.url.url_string,
+            get_dir_path_from_url(&self.url.url_string)
```
vincenzopalazzo, line 117: same there

reckless_lib/src/url.rs
```diff
@@ -0,0 +1,92 @@
+use std::fmt;
+
+use crate::utils::get_reckless_dir;
+
+/// This struct will make sure our URL's are of the
+/// correct format
+#[derive(Clone)]
+pub struct URL {
+    // the url string
```
vincenzopalazzo, line 9:
```suggestion
    /// the url string
```

in this way will be a comment

vincenzopalazzo (changes requested):
Some minor comments, but for me the PR can be merged.

I think just add a FIXME to remember that we can improve the URL implementation

reckless_github/src/lib.rs
swaptr: Fixed. thanks.

reckless_lib/src/url.rs
swaptr: Thanks, just fixed.

reckless_github/src/repository.rs
swaptr: Fixed.

vincenzopalazzo (approved):
Very good cleanup, thanks!

PR #14, opened 2022-11-27
lib: add functionality for install

swaptr:
Add functionality for ``reckless install`.

### DIFF
--- a/reckless_cmd/src/reckless/cmd.rs
+++ b/reckless_cmd/src/reckless/cmd.rs
@@ -19,7 +19,7 @@ pub struct RecklessArgs {
 pub enum RecklessCommand {
     /// Install a single or a list of plugins.
     #[clap(arg_required_else_help = true)]
-    Install,
+    Install { plugin: Vec<String> },
     /// upgrade a single or a list of plugins.
     #[clap(arg_required_else_help = true)]
     Upgrade,
--- a/reckless_cmd/src/reckless/mod.rs
+++ b/reckless_cmd/src/reckless/mod.rs
@@ -46,9 +46,8 @@ impl PluginManager for RecklessManager {
         Ok(())
     }
 
-    async fn install(&mut self, plugins: &[&str]) -> Result<(), RecklessError> {
-        // FIXME: Fix debug message with the list of plugins to be installed
-        debug!("installing plugins");
+    async fn install(&mut self, plugins: Vec<String>) -> Result<(), RecklessError> {
+        debug!("installing plugins {:?}", plugins);
         Ok(())
     }
 
--- a/reckless_cmd/src/reckless/mod.rs
+++ b/reckless_cmd/src/reckless/mod.rs
@@ -17,7 +17,9 @@ mod config;
 
 pub struct RecklessManager {
     config: config::RecklessConf,
-    repos: Vec<Box<dyn Repository>>,
+    /// List of repositories
+    repos: Vec<Box<Github>>,
+    /// List of plugins installed
     plugins: Vec<String>,
 }
 
@@ -48,6 +50,19 @@ impl PluginManager for RecklessManager {
 
     async fn install(&mut self, plugins: &Vec<String>) -> Result<(), RecklessError> {
         debug!("installing plugins {:?}", plugins);
+        for plugin_to_be_installed in plugins {
+            for repo in &self.repos {
+                for plugin_list in repo.list().await {
+                    for mut plugin in plugin_list {
+                        if plugin_to_be_installed == &plugin.name() {
+                            // FIXME: This path needs to be added to config
+                            let path = plugin.configure().await.unwrap();
+                            debug!("{path}");
+                        }
+                    }
+                }
+            }
+        }
         Ok(())
     }
 
--- a/reckless_cmd/src/reckless/mod.rs
+++ b/reckless_cmd/src/reckless/mod.rs
@@ -46,7 +46,7 @@ impl PluginManager for RecklessManager {
         Ok(())
     }
 
-    async fn install(&mut self, plugins: Vec<String>) -> Result<(), RecklessError> {
+    async fn install(&mut self, plugins: &Vec<String>) -> Result<(), RecklessError> {
         debug!("installing plugins {:?}", plugins);
         Ok(())
     }
--- a/reckless_lib/Cargo.toml
+++ b/reckless_lib/Cargo.toml
@@ -10,3 +10,4 @@ git2 = "0.15.0"
 log = "0.4.17"
 env_logger = "0.9.3"
 glob = "0.3.0"
+tokio = { version = "1.22.0", features = ["process"] }
--- a/reckless_lib/src/plugin.rs
+++ b/reckless_lib/src/plugin.rs
@@ -139,6 +139,10 @@ impl Plugin {
     async fn remove(&mut self) -> Result<(), RecklessError> {
         todo!("not implemented yet")
     }
+
+    pub fn name(&self) -> String {
+        self.name.clone()
+    }
 }
 
 impl fmt::Display for Plugin {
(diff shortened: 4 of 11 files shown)


reckless_cmd/src/reckless/cmd.rs
```diff
@@ -19,7 +19,7 @@ pub struct RecklessArgs {
 pub enum RecklessCommand {
     /// Install a single or a list of plugins.
     #[clap(arg_required_else_help = true)]
-    Install { plugin: Vec<String> },
+    Install { plugin_str: Vec<String> },
```
vincenzopalazzo, line 22:
```suggestion
    Install { plugin_str: Vec<String> },
```

why this name? this should be also the name that it is in the command line if you type `cargo run -- --help`

reckless_lib/src/plugin.rs
```diff
@@ -86,7 +88,19 @@ impl Plugin {
             if let Some(script) = &conf.plugin.install {
                 let cmds = script.split("\n"); // Check if the script contains `\`
                 for cmd in cmds {
-                    // FIXME: run command
+                    let cmd_result = Command::new(cmd)
+                        .current_dir(self.path.clone())
+                        .output()
+                        .await;
+                    match cmd_result {
+                        Ok(_) => return Ok(self.path.clone()),
```
vincenzopalazzo, line 96:
```suggestion
                        Ok(_) => {},
```

We need to return the path of the main file and not the root of the plugin, otherwise, we do not solve our problem.

N.B it is already done at line  105

vincenzopalazzo (changes requested):

reckless_lib/Cargo.toml
```diff
@@ -10,3 +10,4 @@ git2 = "0.15.0"
 log = "0.4.17"
 env_logger = "0.9.3"
 glob = "0.3.0"
+tokio = { version = "1.21.2", features = ["full"] }
```
vincenzopalazzo, line 13:
```suggestion
tokio = { version = "1.21.2", features = ["process"] }
```

mh not sure that this is correct, we need only the process feature, you can find the docs here https://docs.rs/tokio/latest/tokio/#feature-flags

vincenzopalazzo (changes requested):

reckless_lib/Cargo.toml
swaptr: Yes, I should have been careful, not just copy blindly :smile:

reckless_lib/src/plugin.rs
```diff
@@ -16,12 +17,24 @@ pub enum PluginLang {
...
         match self {
             PluginLang::Python => {
                 /* 1. RUN PIP install or poetry install
                  * 2. return the path of the main file */
-                todo!()
+                let req_file = format!("{}/requirements.txt", path);
+                let main_file = format!("{}/{}.py", path, name);
+                match Command::new("pip")
+                    .arg("install")
+                    .arg(req_file.as_str())
```
vincenzopalazzo, line 29:
```suggestion
                    .arg("install")
                    .arg("-r")
                    .arg(req_file.as_str())
```

it is missing the `-r`?

reckless_lib/src/plugin.rs
```diff
@@ -16,12 +17,24 @@ pub enum PluginLang {
...
+                    .output()
+                    .await
+                {
+                    Ok(_) => Ok(main_file),
+                    Err(_) => {
+                        return Err(RecklessError::new(1, "problem installing python plugin"))
+                    }
```
vincenzopalazzo, line 36:
```suggestion
                    Err(err) => {
                        return Err(RecklessError::new(1, format!("problem installing python plugin {err}")))
                    }
```

ignoring the error you are missing the reason

reckless_lib/src/plugin.rs
```diff
@@ -86,14 +100,26 @@ impl Plugin {
...
                 let cmds = script.split("\n"); // Check if the script contains `\`
                 for cmd in cmds {
-                    // FIXME: run command
+                    let cmd_result = Command::new(cmd)
+                        .current_dir(self.path.clone())
+                        .output()
+                        .await;
+                    match cmd_result {
+                        Ok(_) => {}
+                        Err(_) => {
```
vincenzopalazzo, line 109: sane here

reckless_lib/src/plugin_manager.rs
```diff
@@ -12,7 +12,7 @@ pub trait PluginManager {
 
     /// install a sequence of plugin or return an error if somethings happens.
     // FIXME: what happens if only one plugin fails?
-    async fn install(&mut self, plugins: &[&str]) -> Result<(), RecklessError>;
+    async fn install(&mut self, plugins: Vec<String>) -> Result<(), RecklessError>;
```
vincenzopalazzo, line 15:
```suggestion
    async fn install(&mut self, plugins: &Vec<String>) -> Result<(), RecklessError>;
```

there is no reason to pass a copy of the value

vincenzopalazzo (changes requested):

reckless_cmd/src/reckless/mod.rs
```diff
@@ -46,9 +48,21 @@ impl PluginManager for RecklessManager {
         Ok(())
     }
 
-    async fn install(&mut self, plugins: &[&str]) -> Result<(), RecklessError> {
-        // FIXME: Fix debug message with the list of plugins to be installed
-        debug!("installing plugins");
+    async fn install(&mut self, plugins: &Vec<String>) -> Result<(), RecklessError> {
+        debug!("installing plugins {:?}", plugins);
+        for plugin_to_be_installed in plugins {
```
vincenzopalazzo, line 53:
```suggestion
        for plugin in plugins {
```

reckless_cmd/src/reckless/mod.rs
```diff
@@ -46,9 +48,21 @@ impl PluginManager for RecklessManager {
...
+            for repo in &self.repos {
+                for plugin_list in repo.list().await {
```
vincenzopalazzo, line 55:
```suggestion
                if let Some(plugin_in_repo) = repo.get_plugin_by_name(plugin_name) {
                      // please avoid unwrapping on configure this is the worst this to do! 
                      // in 50% of the case this method fails, and this will crash reckless
                       let path = plugin_in_repo.configure().await.unwrap();
                       debug!("{path}");
                       continue;
                } 
                // return a not found plugin error
```

You had the power to change the API so we can make the code a little bit cleaner

reckless_cmd/src/reckless/mod.rs
```diff
@@ -46,9 +48,21 @@ impl PluginManager for RecklessManager {
...
+                    for mut plugin in plugin_list {
+                        if plugin_to_be_installed == &plugin.name() {
+                            // FIXME: This path needs to be added to config
+                            let path = plugin.configure().await.unwrap();
+                            debug!("{path}");
+                        }
+                    }
```
vincenzopalazzo, line 62:
```suggestion
```

Source https://www.kernel.org/doc/html/v4.10/process/coding-style.html#indentation

>The answer to that is that if you need more than 3 levels of indentation, you’re screwed anyway, and should fix your program.

reckless_cmd/src/reckless/mod.rs
```diff
@@ -17,7 +17,9 @@ mod config;
 
 pub struct RecklessManager {
     config: config::RecklessConf,
-    repos: Vec<Box<dyn Repository>>,
+    /// List of repositories
+    repos: Vec<Box<Github>>,
```
vincenzopalazzo, line 21: No this is wrong! The previous one was correct!

With your solution, you are breaking all the architecture because we can not work with a repository that it is different from the Github one

vincenzopalazzo (changes requested):
There are good progress but I think this required more love, if you are fiding this hard to understand I can do it

vincenzopalazzo:
Closing in favor of https://github.com/cln-reckless/reckless.rs/pull/18

PR #15, opened 2022-11-27
url: move remote repo to user defined dir

swaptr:
This PR allows cloning remote repo into user defined paths.
Earlier, adding remote repo would result in paths such as `{HOME}/.reckless/repositories/username/repo` which would disregard the `name` argument provided by the user.
Now, adding remote repo would result in path `{HOME}/.reckless/repositories/name/`.


vincenzopalazzo (approved):
I like it! Thanks, it is a very good solution

In addition I just realize that your way to store the default path like `{HOME}/.reckless/repositories/username/repo` allow us to support also firks as different repository! This is clever!

PR #24, opened 2022-12-08
fix: replace walkdir for glob

swaptr:


reckless_github/src/repository.rs
```diff
@@ -57,68 +51,56 @@ impl Github {
...
+                                let conf_path = format!("{}/{}", path_to_plugin, file_name);
+                                let mut conf_str = String::new();
+                                File::open(conf_path.as_str())
+                                    .await?
+                                    .read_to_string(&mut conf_str)
+                                    .await?;
+                                let _: Conf = serde_yaml::from_str(&conf_str).unwrap();
+                                // FIXME: store the conf inside the plugin
+                            }
+                             */
```
swaptr, line 95: I am not sure what needs to be done with this code?

reckless_github/src/repository.rs
```diff
@@ -57,68 +51,56 @@ impl Github {
...
+            .filter_entry(|dir_entry| {
+                dir_entry
+                    .file_name()
+                    .to_str()
+                    .map(|s| {
+                        dir_entry.depth() == 0 || !s.starts_with(".") && dir_entry.path().is_dir()
+                    })
+                    .unwrap()
+            })
+        {
```
vincenzopalazzo, line 68: This looks really unreadable

This is an example from the docs that it is much nicer https://docs.rs/walkdir/latest/walkdir/#example-skip-hidden-files-and-directories-on-unix

reckless_github/src/repository.rs
vincenzopalazzo: This should load the plugin manifest that we encode into yaml format

vincenzopalazzo (changes requested):
much cleaner with this new crate, I think this required some more love but it is good to be merged

reckless_github/src/repository.rs
swaptr: This one is very similar to how we do, and I am presenting a case against using a separate filter function. I have cleaned this up a bit, but if you really want to add a filter function for this, let me know.

vincenzopalazzo: No it is not, you are doing all inside the for and the code example is splitted in function and variable declared outside the for

reckless_github/src/repository.rs
```diff
@@ -43,82 +50,47 @@ impl Github {
...
-                        _ => continue,
+                            "reckless.yml" | "reckless.yaml" => {
+                                let conf_path = format!("{}/{}", path_to_plugin, file_name);
+                                let mut conf_str = String::new();
+                                File::open(conf_path.as_str())
+                                    .await?
+                                    .read_to_string(&mut conf_str)
+                                    .await?;
+                                conf = serde_yaml::from_str(&conf_str).unwrap();
+                                continue;
```
swaptr, line 82: Review on this?

reckless_github/src/repository.rs
```diff
@@ -43,82 +50,47 @@ impl Github {
...
-        //      self.plugins(plugin)
-        for plugin in glob(&pattern).unwrap() {
-            match plugin {
-                Ok(path) => {
-                    let (path_to_plugin, plugin_name) =
-                        get_plugin_info_from_path(path.clone()).unwrap();
+        for plugin_dir in WalkDir::new(repo_path)
+            .max_depth(1)
+            .into_iter()
+            .filter_entry(|dir_entry| !is_hidden(dir_entry))
```
vincenzopalazzo, line 56:
```suggestion
        let target_dirs = WalkDir::new(repo_path)
            .max_depth(1)
            .into_iter()
            .filter_entry(|dir_entry| !is_hidden(dir_entry));
        for plugin_dir in target_dirs
```

reckless_github/src/repository.rs
```diff
@@ -43,82 +50,47 @@ impl Github {
...
-                        }
-                        _ => continue,
+                            "reckless.yml" | "reckless.yaml" => {
+                                let conf_path = format!("{}/{}", path_to_plugin, file_name);
+                                let mut conf_str = String::new();
+                                File::open(conf_path.as_str())
+                                    .await?
+                                    .read_to_string(&mut conf_str)
+                                    .await?;
+                                conf = serde_yaml::from_str(&conf_str).unwrap();
```
vincenzopalazzo, line 81: The conf is decoded in the correct Configuration type?

In addition, can you store the configuration somewhere in the plugin so we can make the PR ready? otherwise this is still not completed

reckless_github/src/repository.rs
```diff
@@ -43,82 +50,47 @@ impl Github {
...
+                                continue;
+                                // FIXME: store the conf inside the plugin
+                            }
+                            _ => continue,
+                        };
+                        debug!("PLUGIN: {} {}", plugin_name, path_to_plugin);
```
vincenzopalazzo, line 87:
```suggestion
                        debug!("new plugion: {} {}", plugin_name, path_to_plugin);
```

reckless_github/src/repository.rs
vincenzopalazzo: we do not need to abuse of a language feature, so I think in this way it is much cleaner

vincenzopalazzo (changes requested):

reckless_github/src/repository.rs
vincenzopalazzo: I suggested a new solution to make the for much readble

reckless_github/src/repository.rs
swaptr: It would seem so.
Also I have stored it [here](https://github.com/cln-reckless/reckless.rs/blob/8bd7a90990f1ec9aac69e39a699b870f2f072c96/reckless_lib/src/plugin.rs#L85)
assigned [here](https://github.com/cln-reckless/reckless.rs/blob/8bd7a90990f1ec9aac69e39a699b870f2f072c96/reckless_lib/src/plugin.rs#L95)

vincenzopalazzo: I do not think that this will work, there is no type in there! but maybe you did not test the code, so for the next time please test it

vincenzopalazzo (changes requested):

PR #27, opened 2022-12-09
coffee: complete code refactoring

swaptr:
code refactoring from reckless -> coffee


vincenzopalazzo (approved):
Thanks, this was the most painful part to do inside the refactoring, thank to do it

PR #29, opened 2022-12-11
feat(storage): implement file storage

vincenzopalazzo:
This is a working-in-progress PR that allows making a file persistence of the plugin manager status.

It is a monkey solution but good enough to unlock some of our blocking tasks

Fixes https://github.com/cln-reckless/coffee/issues/20

Signed-off-by: Vincenzo Palazzo <vincenzopalazzodev@gmail.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -124,6 +124,7 @@ version = "0.1.0"
 dependencies = [
  "async-trait",
  "coffee_lib",
+ "coffee_storage",
  "env_logger",
  "git2",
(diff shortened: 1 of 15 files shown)


vincenzopalazzo:
Ok, I fight with Rust and the architecture of coffe, it required some more love but looks like that it is ready to first land, and we should debug it for a while but I was able to add the remote and try to install a plugin!

Now next step we should link feel the cln conf and link it

PR #48, opened 2023-02-26
Refactor CoffeeCommand::Remove to use plugin parameter

royalpinto007:
This commit refactors the CoffeeCommand::Remove method to accept a plugin parameter instead of using a placeholder todo!() statement. The plugin parameter is then used to call the remove method on the coffee instance, using the await keyword to ensure the operation completes before continuing.


vincenzopalazzo:
from the CI I can guess that you did not compile the code, and I guess that you had not tried to use core lightning and coffee together.

PR #50, opened 2023-02-27
Getting the plugin language from the coffee configuration file

### LINKED ISSUE #49: Inferer the plugin information from the configuration file if exist
# Background

```
cargo run -- setup <path of the core lightning file>
```

Currently, there is a FIXME inside the code https://github.com/coffee-tools/coffee/blob/master/coffee_github/src/repository.rs#L104-L114 that suggest searching before if the plugin has the coffee configuration file, and then guess the language by iterating on each file.

# Goal

So the goal is to refactor the code to get the information from the configuration file, where an example of conf file looks like the following one https://github.com/coffee-tools/coffee/blob/master/coffee.yml

In order to achieve this, you should put create a configuration file inside on of the plugin that are in the repository https://github.com/lightningd/plugins.git and test if your solution wortks

tareknaser360:
# Getting the plugin language from the coffee configuration file

## Description

Currently, the code guesses the plugin language by iterating on each file. This PR makes coffee able to get language information from the coffee configuration file

Fixes #49

## Type of change

- New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- I forked `https://github.com/lightningd/plugins` and added coffee configuration file to the `probe` repo
- I commented out the code which iterates the files to guess the language
- - Then, I setup coffee from source using `cargo run -- remote add lightningd https://github.com/tareknaser360/plugins.git`.
  - output: 
    ```
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
    Running `target/debug/coffee_cmd remote add lightningd 'https://github.com/tareknaser360/plugins.git'`
    ```
- - Without the new code block, running `cargo run -- install probe` yielded 
  - 
    ```
    error[E0425]: cannot find value `conf` in this scope
    warning: `coffee_github` (lib) generated 4 warnings
    error: could not compile `coffee_github` due to previous error; 4 warnings emitted
    ```
- - After adding the code block, running the same command, we get:
  -
    ```
    Compiling coffee_cmd v0.0.1-alpha.1 (/home/tarekbtc/coffee/coffee_cmd)
    Finished dev [unoptimized + debuginfo] target(s) in 3.32s
    Running `target/debug/coffee_cmd install probe`
    ```   

**Test Configuration**:
* Firmware version: Ubuntu 20.04.5 LTS
* Running Bitcoin Core Signet
* Running Core Lightning using btcli4j plugin

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings

### DIFF
--- a/coffee_github/src/repository.rs
+++ b/coffee_github/src/repository.rs
@@ -67,26 +67,6 @@ impl Github {
                     let mut plugin_name = String::new();
                     let mut plugin_lang = PluginLang::Unknown;
 
-                    // try to understand the language from the file
-                    let files = WalkDir::new(plugin_path.path()).max_depth(1);
-                    for file in files {
-                        let file_dir = file.unwrap().clone();
-                        (path_to_plugin, plugin_name) =
-                            get_plugin_info_from_path(file_dir.path()).unwrap();
-
-                        let file_name = file_dir.file_name().to_str().unwrap();
-                        plugin_lang = match file_name {
-                            "requirements.txt" => PluginLang::Python,
-                            "go.mod" => PluginLang::Go,
-                            "cargo.toml" => PluginLang::Rust,
-                            "pubspec.yaml" => PluginLang::Dart,
-                            "package.json" => PluginLang::JavaScript,
-                            "tsconfig.json" => PluginLang::TypeScript,
-                            _ => continue,
-                        };
-                    }
-                    debug!("possible plugin language: {:?}", plugin_lang);
-
                     // check if the plugin has the custom configuration to read.
                     let mut conf = None;
                     for file in ["coffee.yaml", "coffee.yml"] {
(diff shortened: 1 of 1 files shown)


coffee_github/src/repository.rs
```diff
@@ -97,14 +97,20 @@ impl Github {
...
+                            let conf_lang = (&conf_file.plugin.lang).to_owned();
+                            plugin_lang = match conf_lang.as_str() {
+                                "py" => PluginLang::Python,
+                                "go" => PluginLang::Go,
+                                "rs" => PluginLang::Rust,
+                                "dart" => PluginLang::Dart,
+                                "js" => PluginLang::JavaScript,
+                                "ts" => PluginLang::TypeScript,
+                                _ => continue,
+                            };
```
vincenzopalazzo, line 109: nice, this is was I was thinking, but in terms of performance this isn't good.

If you look at all the methods, before finding the configuration file we try to guess what kind of language is by iterating across all the plugin files, this is an O(file_num) operation that we can skip by looking before at the conf file and then if the conf file is missing, try to iterate across all the plugin file and guess what the language is.

What do you think?

coffee_github/src/repository.rs
```diff
@@ -97,14 +97,20 @@ impl Github {
...
                             let conf_file = serde_yaml::from_str::<Conf>(&conf_str).unwrap();
+                            let conf_lang = (&conf_file.plugin.lang).to_owned();
+                            plugin_lang = match conf_lang.as_str() {
+                                "py" => PluginLang::Python,
+                                "go" => PluginLang::Go,
+                                "rs" => PluginLang::Rust,
+                                "dart" => PluginLang::Dart,
+                                "js" => PluginLang::JavaScript,
+                                "ts" => PluginLang::TypeScript,
+                                _ => continue,
```
vincenzopalazzo, line 108:
```suggestion
                                _ => CoffeError::new(1, format!("language {conf_lang} not supported"),
```

This is not a continue, but this is an error because this mean that we do not know the language

vincenzopalazzo (changes requested):
Nice PR description, I left a comment because I think we can do better in the code

coffee_github/src/repository.rs
tareknaser360: I refactored the match expression to be able to return a `CoffeeError`.

coffee_github/src/repository.rs
```diff
@@ -98,18 +78,42 @@ impl Github {
...
+
+                            let file_name = file_dir.file_name().to_str().unwrap();
+                            plugin_lang = match file_name {
+                                "requirements.txt" => PluginLang::Python,
+                                "go.mod" => PluginLang::Go,
+                                "cargo.toml" => PluginLang::Rust,
+                                "pubspec.yaml" => PluginLang::Dart,
+                                "package.json" => PluginLang::JavaScript,
+                                "tsconfig.json" => PluginLang::TypeScript,
+                                _ => continue,
```
vincenzopalazzo, line 112:
```suggestion
                                _ => PluginLang::Unknown,
```

Should be available this, right?

vincenzopalazzo (commented):
just small nit and also I think this required some `make` command to make happy the CI

vincenzopalazzo (dismissed):
Ok looks good!

Before merging it you should fix the commit history because not it is a little bit messy

In this case you can just squash the commit in a single one?

tareknaser360:
> Before merging it you should fix the commit history because not it is a little bit messy
> 
> In this case you can just squash the commit in a single one?

Done

vincenzopalazzo (approved):

PR #62, opened 2023-03-07
Create core lightning config file if missing

### LINKED ISSUE #57: allow to create a empty core lightning config when it is missed inside the `./lightning/<network>/`
# Background

```
cargo run -- setup <path of the core lightning file>
```

Currently, if the user does have not the config file, the user should create an empty file, this could be done by coffee and it is already supported by the design of our configuration parser.

# Goal

the following command should create the empty configuration file if it is missed

```
cargo run -- setup <path of the core lightning file>
```

tareknaser360:
# Description
This option is already implemented in CLNConf Parser.
`pub(crate) fn new(file_path: &str, create_if_missing: bool)`

I just set `create_if_missing` to be true.

Fixes #57

### DIFF
--- a/coffee_cmd/src/coffee/mod.rs
+++ b/coffee_cmd/src/coffee/mod.rs
@@ -117,7 +117,7 @@ impl CoffeeManager {
             return Ok(());
         }
         let path = self.config.cln_config_path.clone().unwrap();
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
Looks like you are really having fun

Issue #13, opened 2022-11-25
support for other commands


swaptr:
the following command needs to be supported too

- [ ] reckless upgrade (https://github.com/coffee-tools/coffee/issues/43)
- [x] reckless list (https://github.com/coffee-tools/coffee/issues/42)
- [ ] reckless remove (https://github.com/coffee-tools/coffee/issues/41)
- [x] remove the remote added (https://github.com/coffee-tools/coffee/pull/80)

tareknaser360:
Done with #41 and #132

Issue #41, opened 2023-02-24
Support `remove` command to remove a plugin that is currently installed by coffe


vincenzopalazzo:
# Background

To run Coffe you are run the following command `cargo run -- --help` and you can run the following command to setup coffee for your core lightning node

```
cargo run -- setup <core lightning root path>
cargo run -- remote add lightningd https://github.com/lightningd/plugins.git
cargo run -- install <plugin_name>
```

Currently, there is no way to remove the plugin that we just installed, so this task required you to implement the remove command.

# Goal

Be able to run the following command 
```
cargo run -- remove <name_plugin>
```

royalpinto007:
Hey @vincenzopalazzo,
Please assign this issue to me, and I will start working on it right away.

tareknaser360:
I was planning to work on this issue. Is there any pre-SoB issues I can work on?

royalpinto007:
As I believe changes are required in three files, I will be working on this issue in steps by adding commits to each file one by one.

royalpinto007:
https://github.com/coffee-tools/coffee/pull/48

This improves the functionality of the CoffeeCommand::Remove method by allowing the removal of specific plugins from the coffee instance. This change also improves the overall maintainability of the code by replacing the placeholder statement with a functional implementation.

I will do the changes in the remaining files too.

vasvigarg:
is the issue resolved yet or can i work on it?

vincenzopalazzo:
@vasvigarg you can!

vasvigarg:
pls assign

vincenzopalazzo:
as you can see there is no need to assign!

PR #80, opened 2023-03-17
feat(core): add remove remote command

tareknaser360:
running `cargo run -- remote remove <repo name>` will: 
- remove the directory from local storage
- update the JSON file with the dump of coffee status

This is related to #13

### DIFF
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -20,6 +20,7 @@ use serde::{Deserialize, Serialize};
 use serde_json::json;
 use serde_json::Value;
 use std::fmt::Debug;
+use std::fs;
 use std::vec::Vec;
 
 use super::config;
@@ -278,6 +279,21 @@ impl PluginManager for CoffeeManager {
         repo.init().await?;
         self.repos.push(Box::new(repo));
         debug!("remote added: {} {}", name, &url.url_string);
+        self.storage.store(&self.storage_info()).await?;
+        Ok(())
+    }
+
+    async fn rm_remote(&mut self, name: &str) -> Result<(), CoffeeError> {
+        debug!("remote removing: {}", name);
+        let index = self
+            .repos
+            .iter()
+            .position(|x| &*x.to_owned().name() == name)
+            .unwrap();
+        let repo_path = &self.repos[index].url().path_string;
+        fs::remove_dir_all(repo_path)?;
+        self.repos.remove(index);
+        debug!("remote removed: {}", name);
         self.storage.store(&self.storage_info()).await?;
         Ok(())
     }
--- a/coffee_lib/src/plugin_manager.rs
+++ b/coffee_lib/src/plugin_manager.rs
@@ -28,6 +28,9 @@ pub trait PluginManager {
     /// add the remote repository to the plugin manager.
     async fn add_remote(&mut self, name: &str, url: &str) -> Result<(), CoffeeError>;
 
+    /// remove the remote repository from the plugin manager.
+    async fn rm_remote(&mut self, name: &str) -> Result<(), CoffeeError>;
+
     /// set up the core lightning configuration target for the
     /// plugin manager.
     async fn setup(&mut self, cln_conf_path: &str) -> Result<(), CoffeeError>;
--- a/docs/docs-book/src/using-coffee.md
+++ b/docs/docs-book/src/using-coffee.md
@@ -54,10 +54,10 @@ coffee remote add <name repository> <url repository>
 
 To remove a plugin repository, simply run the following command.
 
-> ⚠️  this feature is under development, see [the tracking issue](https://github.com/coffee-tools/coffee/issues/13)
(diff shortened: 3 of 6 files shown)


coffee_core/src/coffee.rs
```diff
@@ -281,6 +282,21 @@ impl PluginManager for CoffeeManager {
...
+            .iter()
+            .position(|x| &*x.to_owned().name() == name)
+            .unwrap();
+        let repo_path = &self.repos[index].url().path_string;
+        fs::remove_dir_all(repo_path).unwrap();
+        self.repos.remove(index);
+        debug!("remote removed: {}", name);
+        self.storage.store(&self.storage_info()).await?;
+        Ok(())
+    }
```
vincenzopalazzo, line 299: This is not completed, because if you remove the directory as you do we need to remove also the plugin (if any)

After I removed the repository the plugin installed are still there

```
➜  coffee git:(pr80) cargo run -- --network testnet list --remotes 
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/coffee --network testnet list --remotes`
{
  "plugins": [
    {
      "conf": {
        "plugin": {
          "dependencies": null,
          "deprecated": null,
          "install": "sh -C ./gradlew createRunnableScript\n",
          "lang": "java",
          "main": "btcli4j-gen.sh",
          "name": "btcli4j",
          "version": "0.0.1"
        }
      },
      "lang": "JVM",
      "name": "btcli4j",
      "path": "/home/vincent/.coffee/repositories/lightnind/btcli4j",
      "root_path": "/home/vincent/.coffee/repositories/lightnind/btcli4j"
    },
    {
      "conf": null,
      "lang": "Python",
      "name": "summary",
      "path": "/home/vincent/.coffee/repositories/lightnind/summary",
      "root_path": "/home/vincent/.coffee/repositories/lightnind/summary"
    }
  ],
  "remotes": []
}

```

coffee_core/src/coffee.rs
```diff
@@ -281,6 +282,21 @@ impl PluginManager for CoffeeManager {
...
+
+    async fn remove_remote(&mut self, name: &str) -> Result<(), CoffeeError> {
+        debug!("remote removing: {}", name);
+        let index = self
+            .repos
+            .iter()
+            .position(|x| &*x.to_owned().name() == name)
+            .unwrap();
+        let repo_path = &self.repos[index].url().path_string;
+        fs::remove_dir_all(repo_path).unwrap();
```
vincenzopalazzo, line 294:
```suggestion
        fs::remove_dir_all(repo_path)?;
```

or you can use the `map_err(|err| CoffeError::new(1, &format!("{err}"))?`

The unwrap should be never called inside a function call because you are breaking the normal life cycle of an application.

coffee_lib/src/plugin_manager.rs
```diff
@@ -28,6 +28,9 @@ pub trait PluginManager {
     /// add the remote repository to the plugin manager.
     async fn add_remote(&mut self, name: &str, url: &str) -> Result<(), CoffeeError>;
 
+    /// remove the remote repository from the plugin manager.
+    async fn remove_remote(&mut self, name: &str) -> Result<(), CoffeeError>;
```
vincenzopalazzo, line 32:
```suggestion
    async fn rm_remote(&mut self, name: &str) -> Result<(), CoffeeError>;
```

docs/docs-book/src/using-coffee.md
```diff
@@ -54,7 +54,7 @@ coffee remote add <name repository> <url repository>
 
 To remove a plugin repository, simply run the following command.
 
-> ⚠️  this feature is under development, see [the tracking issue](https://github.com/coffee-tools/coffee/issues/13)
+> ✅ Implemented
 
 ```bash
 coffee remote remove <NAME_OF_THE_REPOSITORY>
```
vincenzopalazzo, line 60:
```suggestion
coffee remote rm <NAME_OF_THE_REPOSITORY>
```

What do you think to rename the remove to rm? looks like more readble to me

vincenzopalazzo (changes requested):
OK some consideration, I think removing the plugin should be done after this https://github.com/coffee-tools/coffee/issues/41 is implemented

In this way, you can reuse the `rm_plugin` function

BTW the PR is good enough so we make the rename and the error things and we can merge it.

In the meanwhile can you open an issue to point out that this PR remove the repository without removing the plugin? So we can avoid to forget it

coffee_core/src/coffee.rs
tareknaser360: Will the `rm_plugin` function take just the name of the plugin?

I can make an empty function and call it on the plugins installed in the remote repo so that it's automatically used when we implement it.

vincenzopalazzo: lets wait for the moment :) but guest the rm take the name of the plugin ony

vincenzopalazzo (approved):

PR #82, opened 2023-03-18
fix: Update installation command for btcli4j

royalpinto007:
The current installation command for btcli4j on the testnet network is incorrect and does not work as expected. This commit updates the command to use 'coffee --network testnet install btcli4j' which installs the package correctly.


vincenzopalazzo (approved):
Thanks, the commit header is wrong, it needs to be `fix(docs): ....`

We are not ready to generate the changelog but start to pay attention to it for simple PR like this avoid to Following the guideline can result in rejecting the PR we have enough docs to avoid this mistake

cc @royalpinto007

PR #85, opened 2023-03-21
feat(core): add show command

### LINKED ISSUE #13: support for other commands

tareknaser360:
Fixes #13

### DIFF
--- a/coffee_cmd/src/main.rs
+++ b/coffee_cmd/src/main.rs
@@ -8,6 +8,9 @@ use coffee_core::coffee::CoffeeManager;
 use coffee_lib::errors::CoffeeError;
 use coffee_lib::plugin_manager::PluginManager;
 
+use termimad::crossterm::style::Color::*;
+use termimad::{Alignment, MadSkin};
+
 #[tokio::main]
 async fn main() -> Result<(), CoffeeError> {
     env_logger::init();
@@ -42,6 +45,21 @@ async fn main() -> Result<(), CoffeeError> {
             // and the coffee script
             coffee.setup(&cln_conf).await
         }
+        CoffeeCommand::Show { plugin } => match coffee.show(&plugin).await {
+            Ok(val) => {
+                let mut skin = MadSkin::default();
+                skin.table.align = Alignment::Center;
+                skin.set_headers_fg(AnsiValue(178));
+                skin.bold.set_fg(Yellow);
+                skin.italic.set_fg(Magenta);
+                skin.scrollbar.thumb.set_fg(AnsiValue(178));
+                skin.code_block.align = Alignment::Center;
+
+                skin.print_text(val["show"].as_str().unwrap());
+                Ok(())
+            }
+            Err(err) => Err(err),
+        },
     };
 
     if let Err(err) = result {
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -297,6 +297,21 @@ impl PluginManager for CoffeeManager {
         self.storage.store(&self.storage_info()).await?;
         Ok(())
     }
+
+    async fn show(&mut self, plugin: &str) -> Result<Value, CoffeeError> {
+        for repo in &self.repos {
+            if let Some(plugin) = repo.get_plugin_by_name(plugin) {
+                let readme_path = format!("{}/README.md", plugin.path);
+                let contents = fs::read_to_string(readme_path)?;
+                return Ok(json!({ "show": contents }));
+            }
+        }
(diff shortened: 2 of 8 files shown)


coffee_core/src/coffee.rs
```diff
@@ -297,6 +297,21 @@ impl PluginManager for CoffeeManager {
         self.storage.store(&self.storage_info()).await?;
         Ok(())
     }
+
+    async fn show(&mut self, plugin: &str) -> Result<String, CoffeeError> {
```
vincenzopalazzo, line 301:
```suggestion
    async fn show(&mut self, plugin: &str) -> Result<Value, CoffeeError> {
```

coffee_core/src/coffee.rs
```diff
@@ -297,6 +297,21 @@ impl PluginManager for CoffeeManager {
...
+        for repo in &self.repos {
+            if let Some(plugin) = repo.get_plugin_by_name(plugin) {
+                let readme_path = format!("{}/README.md", plugin.path);
+                let contents = fs::read_to_string(readme_path)?;
+                return Ok(contents);
```
vincenzopalazzo, line 306:
```suggestion
                return Ok(json!{ "show": contents });
```

The core need to return always a json otherwise you return different kind of data for different command and this make the code a mess!

coffee_lib/src/plugin_manager.rs
```diff
@@ -34,4 +34,7 @@ pub trait PluginManager {
     /// set up the core lightning configuration target for the
     /// plugin manager.
     async fn setup(&mut self, cln_conf_path: &str) -> Result<(), CoffeeError>;
+
+    /// show the README file of the pulgin
+    async fn show(&mut self, plugin: &str) -> Result<String, CoffeeError>;
```
vincenzopalazzo, line 39: Same 
```suggestion
    async fn show(&mut self, plugin: &str) -> Result<Value, CoffeeError>;
```

vincenzopalazzo (changes requested):

coffee_core/src/coffee.rs
tareknaser360: Thanks for pointing that out

coffee_core/src/coffee.rs
```diff
@@ -297,6 +297,21 @@ impl PluginManager for CoffeeManager {
         self.storage.store(&self.storage_info()).await?;
         Ok(())
     }
+
+    async fn show(&mut self, plugin: &str) -> Result<Value, CoffeeError> {
+        for repo in &self.repos {
+            if let Some(plugin) = repo.get_plugin_by_name(plugin) {
+                let readme_path = format!("{}/README.md", plugin.path);
```
vincenzopalazzo, line 304:
```suggestion
                let readme_path = format!("{}/README.md", plugin.path);
```

we should also consider other options here like README only, but we can do later

coffee_cmd/src/main.rs
```diff
@@ -42,6 +45,21 @@ async fn main() -> Result<(), CoffeeError> {
...
+                skin.set_headers_fg(AnsiValue(178));
+                skin.bold.set_fg(Yellow);
+                skin.italic.set_fg(Magenta);
+                skin.scrollbar.thumb.set_fg(AnsiValue(178));
+                skin.code_block.align = Alignment::Center;
+
+                skin.print_text(val["show"].as_str().unwrap());
+                Ok(())
+            }
+            Err(err) => Err(err),
```
vincenzopalazzo, line 61: This is nice, thank you!

I was wondering if we can do a little bit better and allow to scroll the terminal and not just print all the content?

Here is an example https://github.com/Canop/termimad/blob/main/examples/scrollable/main.rs

vincenzopalazzo (approved):
We can address the following comments later.

But please open an issue to keep track of it.

BTW nice job @tareknaser360

coffee_core/src/coffee.rs
tareknaser360: Yes sure. Is there any other option other than `README` and `README.md`?

vincenzopalazzo: I think this is the best place to put a `// FIXME: there are more README file options?`

PR #93, opened 2023-04-12
Updated coffee.rs

itschiggy:
fix(core): typo fix in error message in the coffee manager

change "this should never happens" to "this should never happen"

### DIFF
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -42,7 +42,7 @@ impl From<&CoffeeManager> for CoffeStorageInfo {
             let repo = if let Some(git) = repo.as_any().downcast_ref::<Github>() {
                 RepositoryInfo::from(git)
             } else {
-                panic!("this should never happens")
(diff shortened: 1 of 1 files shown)


vincenzopalazzo:
Thanks, please modify the commit header according to this https://coffee-docs.netlify.app/contributing-to-coffee.html

itschiggy:
Modified with the same, please review

vincenzopalazzo:
>Modified with the same, please review

Modify with the same meaning that you did not modify anything! Please read the contribution guideline

itschiggy:
Sorry mate, it's my first time contributing to a project, guess I modified a comment instead of the actual commit message. I searched a lot on how to edit the actual message, and it looks like the only option was to force push, which is said to make the maintainers job harder, so a sure shot no-no. I guess I'll just leave it here for now and raise a different PR fixing this when I'm done with adding a command I've been working on for coffee.
Thanks!

PR #99, opened 2023-04-14
feat(core): add search command

### LINKED ISSUE #98: search command

tareknaser360:
fixes #98

### DIFF
--- a/coffee_cmd/src/main.rs
+++ b/coffee_cmd/src/main.rs
@@ -115,6 +115,14 @@ async fn main() -> Result<(), CoffeeError> {
             }
             Err(err) => Err(err),
         },
+        CoffeeCommand::Search { plugin } => match coffee.search(&plugin).await {
+            Ok(val) => {
+                let repository_url = val.repository_url.as_str();
+                term::success!("found plugin {plugin} in remote repository {repository_url}");
+                Ok(())
+            }
+            Err(err) => Err(err),
+        },
         CoffeeCommand::Nurse {} => {
             term::info!("Nurse command is not implemented");
             Ok(())
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -379,6 +379,19 @@ impl PluginManager for CoffeeManager {
         Err(err)
     }
 
+    async fn search(&mut self, plugin: &str) -> Result<CoffeeSearch, CoffeeError> {
+        for repo in self.repos.values() {
+            if let Some(plugin) = repo.get_plugin_by_name(plugin) {
+                return Ok(CoffeeSearch {
+                    repository_url: repo.url().url_string,
+                    plugin,
+                });
+            }
+        }
+        let err = CoffeeError::new(404, &format!("unable to locate plugin `{plugin}`"));
+        Err(err)
+    }
+
     async fn nurse(&mut self) -> Result<(), CoffeeError> {
         unimplemented!("nurse command is not implemented")
     }
--- a/coffee_lib/src/plugin_manager.rs
+++ b/coffee_lib/src/plugin_manager.rs
@@ -43,6 +43,9 @@ pub trait PluginManager {
     /// show the README file of the plugin
     async fn show(&mut self, plugin: &str) -> Result<CoffeeShow, CoffeeError>;
 
+    /// search remote repositories for a plugin by name
+    async fn search(&mut self, plugin: &str) -> Result<CoffeeSearch, CoffeeError>;
+
     /// clean up storage information about the remote repositories of the plugin manager.
     async fn nurse(&mut self) -> Result<(), CoffeeError>;
 }
--- a/docs/docs-book/src/using-coffee.md
+++ b/docs/docs-book/src/using-coffee.md
@@ -130,6 +130,14 @@ coffee list
 ```bash
 coffee show <plugin_name>
 ```
+
+## Searching for a plugin in remote repositories 
+
+> ✅ Implemented
+
+```bash
+coffee search <plugin_name>
+```
 _________
 ## Running coffee as a server
 
--- a/coffee_cmd/src/cmd.rs
+++ b/coffee_cmd/src/cmd.rs
@@ -50,6 +50,9 @@ pub enum CoffeeCommand {
     /// show the README file of the plugin
     #[clap(arg_required_else_help = true)]
     Show { plugin: String },
+    /// search the remote repositories for a plugin
+    #[clap(arg_required_else_help = true)]
+    Search { plugin: String },
     /// clean up remote repositories storage information
     #[clap(arg_required_else_help = false)]
     Nurse {},
@@ -76,6 +79,7 @@ impl From<&CoffeeCommand> for coffee_core::CoffeeOperation {
             CoffeeCommand::Remote { action } => Self::Remote(action.into()),
             CoffeeCommand::Remove { plugin } => Self::Remove(plugin.to_owned()),
             CoffeeCommand::Show { plugin } => Self::Show(plugin.to_owned()),
+            CoffeeCommand::Search { plugin } => Self::Search(plugin.to_owned()),
             CoffeeCommand::Nurse {} => Self::Nurse,
         }
     }
--- a/coffee_core/src/lib.rs
+++ b/coffee_core/src/lib.rs
@@ -17,6 +17,8 @@ pub enum CoffeeOperation {
     /// Setup(core lightning root path)
     Setup(String),
     Show(String),
+    /// Search(plugin name)
+    Search(String),
     Nurse,
 }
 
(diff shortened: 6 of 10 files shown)


coffee_cmd/src/main.rs
```diff
@@ -78,6 +78,19 @@ async fn main() -> Result<(), CoffeeError> {
             }
             Err(err) => Err(err),
         },
+        CoffeeCommand::Search { plugin } => match coffee.search(&plugin).await {
+            Ok(val) => {
+                let val = val["search"].as_str().unwrap();
+                if val == "notfound" {
+                    term::markdown(&("unable to locate plugin ".to_owned() + &plugin));
```
vincenzopalazzo, line 85: `term::markdown` is wrong!

There are `term::success` and other calls, please look at the code

vincenzopalazzo, line 85: No I do not like this! You can return an error here!

vincenzopalazzo (changes requested):
Some initial comments

coffee_cmd/src/main.rs
tareknaser360: Yes. I modified the code to use these calls

tareknaser360: Does the changes I made suffice or should the function print the coffee error itself?

vincenzopalazzo: The error is print at the line 97

coffee_cmd/src/main.rs
```diff
@@ -78,6 +78,19 @@ async fn main() -> Result<(), CoffeeError> {
             }
             Err(err) => Err(err),
         },
+        CoffeeCommand::Search { plugin } => match coffee.search(&plugin).await {
+            Ok(val) => {
+                let val = val["search"].as_str().unwrap();
+                if val == "not found" {
+                    term::error(&("unable to locate plugin ".to_owned() + &plugin));
```
vincenzopalazzo, line 85:
```suggestion
                    term::error(&format!("unable to locate `{plugin}` plugin "));
```

coffee_cmd/src/main.rs
```diff
@@ -78,6 +78,19 @@ async fn main() -> Result<(), CoffeeError> {
...
+                    Ok(())
+                } else {
+                    term::success!("found plugin {} in remote repository {}", plugin, val);
```
vincenzopalazzo, line 88:
```suggestion
                    term::success!("found plugin {plugin} in remote repository {val}");
```

coffee_core/src/coffee.rs
```diff
@@ -330,6 +330,15 @@ impl PluginManager for CoffeeManager {
         );
         Err(err)
     }
+
+    async fn search(&mut self, plugin: &str) -> Result<Value, CoffeeError> {
+        for repo in &self.repos {
+            if let Some(_plugin) = repo.get_plugin_by_name(plugin) {
+                return Ok(json!({ "search": repo.name() }));
```
vincenzopalazzo, line 337:
```suggestion
                return Ok(json!({ "search": repo.name() }));
```

I would like put a more complex json here, like 

```json
{
   "repository":  "<link>",
  "plugin": {
          "conf": {
            "plugin": {
              "dependencies": null,
              "deprecated": null,
              "install": "make dep\nmake coffee\n",
              "lang": "go",
              "main": "go-lnmetrics",
              "name": "lnmetrics",
              "version": "0.0.6-rc2"
            }
          },
          "lang": "Go",
          "name": "lnmetrics",
          "path": "/home/vincent/.coffee/repositories/lightningd/go-lnmetrics.reporter",
          "root_path": "/home/vincent/.coffee/repositories/lightningd/go-lnmetrics.reporter"
        }
}
```

Inside the plugin field, we store the same information that are inside the `coffee list` command

coffee_core/src/coffee.rs
```diff
@@ -330,6 +330,15 @@ impl PluginManager for CoffeeManager {
...
+            }
+        }
+        return Ok(json!({ "search": "not found" }));
```
vincenzopalazzo, line 340:
```suggestion
        return Err(CoffeErrro::new(404, format!("plugin `{plugin}` not found")))
```

Still wrong, this is the correct way to do it

docs/docs-book/src/using-coffee.md
```diff
@@ -129,6 +129,14 @@ or to see all the plugins inside the repositories
 coffee list --remotes
 ```
 
+## Searching for a plugin
+
+> ✅ Implemented
+
+```bash
+coffee search <NAME_OF_PLUGIN>
```
vincenzopalazzo, line 137:
```suggestion
coffee search <plugin_name>
```

Do not use capitalize letter

coffee_cmd/src/main.rs
```diff
@@ -78,6 +78,19 @@ async fn main() -> Result<(), CoffeeError> {
             }
             Err(err) => Err(err),
         },
+        CoffeeCommand::Search { plugin } => match coffee.search(&plugin).await {
+            Ok(val) => {
+                let val = val["search"].as_str().unwrap();
+                if val == "not found" {
```
vincenzopalazzo, line 84: This is a very wrong thing to do in rust! You had Option types for this reason

coffee_cmd/src/main.rs
vincenzopalazzo, line 85: this part of the if needs to be removed! The error is returned out of the match and then print on the terminal at the line 97

vincenzopalazzo (changes requested):

coffee_cmd/src/main.rs
vincenzopalazzo: yes

coffee_core/src/coffee.rs
tareknaser360: Alright. So this should be the returned JSON but the messages should stay the same?
if found: ✓ found plugin helpme in remote repository https://github.com/lightningd/plugins

docs/docs-book/src/using-coffee.md
tareknaser360: should I change all the docs to use plugin_name instead of NAME_OF_PLUGIN?

coffee_cmd/src/main.rs
tareknaser360: Yes thanks for pointing that out

coffee_core/src/coffee.rs
vincenzopalazzo: we will fix the message later, for now any kind of message is fine, so yes let keep this as it is

docs/docs-book/src/using-coffee.md
vincenzopalazzo: yes but not in this PR

coffee_core/src/coffee.rs
```diff
@@ -330,6 +330,19 @@ impl PluginManager for CoffeeManager {
         );
         Err(err)
     }
+
+    async fn search(&mut self, plugin: &str) -> Result<Value, CoffeeError> {
+        for repo in &self.repos {
+            if let Some(plugin) = repo.get_plugin_by_name(plugin) {
+                return Ok(json!({
+                    "repository": repo.url().url_string,
+                    "plugin": serde_json::to_value(&plugin).unwrap(),
```
vincenzopalazzo, line 339:
```suggestion
                    "plugin": serde_json::to_value(&plugin)?,
```

Most of the time call unwrap is wrong because will panic the program

vincenzopalazzo (changes requested):

coffee_core/src/coffee.rs
tareknaser360: Alright.
I had to map the error to CoffeeError

coffee_lib/src/plugin_manager.rs
```diff
@@ -40,4 +40,7 @@ pub trait PluginManager {
 
     /// show the README file of the pulgin
     async fn show(&mut self, plugin: &str) -> Result<Value, CoffeeError>;
+
+    /// search the remote repositories for the plugin
+    async fn search(&mut self, plugin: &str) -> Result<Value, CoffeeError>;
```
vincenzopalazzo, line 45: Ok, wait a minute here, now I was thinking what is the meaning of this `coffee seach` command?

What is the meaning of knowing the remote if the user adds the remote after all?

vincenzopalazzo (commented):
Ok, the code looks good, but now lets talk about the meaning of this command, I am not sure that this command make sense

coffee_lib/src/plugin_manager.rs
tareknaser360: In my opinion, the purpose of this command would be to make sure whether the plugin is present in the added remote repositories or not, which can be useful prior to installation. 

Nonetheless, after further consideration, it appears that the `coffee show` could fulfill the same objective in a different way.

vincenzopalazzo: I really do not know if this command is the right things to add, reckless does this because it has an internal repository, but we use the remote, so you should know already that this plugin is already in the repository.

Lets leave this here for a while and work on other commands that we needs more than this, like removing a plugin

vincenzopalazzo (approved):
LGTM

Thanks

PR #106, opened 2023-04-21
fix(git): fix derivation of the exec path 

vincenzopalazzo:
for description see commits

### DIFF
--- a/coffee_lib/src/plugin.rs
+++ b/coffee_lib/src/plugin.rs
@@ -81,9 +81,9 @@ impl PluginLang {
 pub struct Plugin {
     name: String,
     /// root path of the plugin
-    root_path: String,
+    pub root_path: String,
     /// path of the main file
-    pub path: String,
+    pub exec_path: String,
     lang: PluginLang,
     conf: Option<Conf>,
 }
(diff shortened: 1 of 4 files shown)


tareknaser360 (dismissed):
This looks good to me.
Thanks

coffee_lib/src/plugin.rs
tareknaser360, line 1: `default_install` stills builds the `exec_path` using string literals 
https://github.com/coffee-tools/coffee/blob/2420a931db446c89094e395c443420b27132fd4a/coffee_lib/src/plugin.rs#L41

This is equivalent to `exec_path`

coffee_lib/src/plugin.rs
```diff
@@ -113,15 +113,15 @@ impl Plugin {
...
             } else {
                 self.lang
-                    .default_install(&self.path, &self.name, verbose)
+                    .default_install(&self.root_path, &self.name, verbose)
                     .await?
             }
         } else {
             self.lang
-                .default_install(&self.path, &self.name, verbose)
+                .default_install(&self.exec_path, &self.name, verbose)
```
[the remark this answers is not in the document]
vincenzopalazzo: yeah this is the root path, food catch!

coffee_lib/src/plugin.rs
vincenzopalazzo: yes but in the PluginLang we do not have the plugin path right? and we need to have the root path because we need to run the command in some way

PR #112, opened 2023-04-23
fix(core): delete installed plugins with removing repository

### LINKED ISSUE #81: remove installed plugins from a remote repository when it's removed
This is to be done after #41 is fixed

tareknaser360:
Fixes #81

The added code simply calls `remove` on plugins that are mutual between the installed plugins and the list of the plugins of the repository being removed


vincenzopalazzo (approved):
Yeah this make a lot of sense, otherwise we leave zombie plugins

Good catch

PR #114, opened 2023-04-23
fix warnings that cargo clippy outputs

### LINKED ISSUE #109: fix cargo clippy
I never test the code with ` cargo clippy --all --tests`

We should fix the warnings :)

tareknaser360:
fixes #109


vincenzopalazzo (approved):

coffee_plugin/src/plugin.rs
```diff
@@ -1,5 +1,5 @@
-pub mod plugin;
+pub mod plugin_mod;
 mod state;
 
-pub use plugin::*;
+pub use plugin_mod::*;
```
vincenzopalazzo, line 4: I do not like this, I preferer on top of this the `#[allow(....)]`

PR #115, opened 2023-04-26
fix(coffee): don't allow duplicate repositories 

tareknaser360:
- added a hashset of the remote repositories names to the storage information of coffee on disk (JSON)
- added a function to check if the remote repository clone was deleted from disk manually and called it when needed

### DIFF
--- a/coffee_cmd/src/cmd.rs
+++ b/coffee_cmd/src/cmd.rs
@@ -50,6 +50,9 @@ pub enum CoffeeCommand {
     /// show the README file of the plugin
     #[clap(arg_required_else_help = true)]
     Show { plugin: String },
+    /// clean up remote repositories storage information
+    #[clap(arg_required_else_help = false)]
+    Nurse {},
 }
 
 #[derive(Debug, Subcommand)]
@@ -73,6 +76,7 @@ impl From<&CoffeeCommand> for coffee_core::CoffeeOperation {
             CoffeeCommand::Remote { action } => Self::Remote(action.into()),
             CoffeeCommand::Remove { plugin } => Self::Remove(plugin.to_owned()),
             CoffeeCommand::Show { plugin } => Self::Show(plugin.to_owned()),
+            CoffeeCommand::Nurse {} => Self::Nurse,
         }
     }
 }
--- a/coffee_cmd/src/main.rs
+++ b/coffee_cmd/src/main.rs
@@ -7,6 +7,7 @@ use radicle_term as term;
 use coffee_core::coffee::CoffeeManager;
 use coffee_lib::errors::CoffeeError;
 use coffee_lib::plugin_manager::PluginManager;
+use coffee_lib::types::NurseStatus;
 
 use crate::cmd::CoffeeArgs;
 use crate::cmd::CoffeeCommand;
@@ -98,6 +99,22 @@ async fn main() -> Result<(), CoffeeError> {
             }
             Err(err) => Err(err),
         },
+        CoffeeCommand::Nurse {} => {
+            let mut spinner =
+                term::spinner("Trying restoring from a corrupted status!".to_string());
+            match coffee.nurse().await {
+                Ok(val) => {
+                    match val.status {
+                        NurseStatus::Corrupted => spinner.message("Storage refreshed"),
+                        NurseStatus::Sane => spinner
+                            .message("Storage files are not corrupt. No need to run coffee nurse"),
+                    }
+                    spinner.finish();
+                }
+                Err(err) => spinner.error(format!("Error while refreshing storage: {err}")),
+            };
+            Ok(())
+        }
     };
 
     if let Err(err) = result {
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -1,7 +1,9 @@
 //! Coffee mod implementation
+use std::collections::HashMap;
 use std::fmt::Debug;
-use std::fs;
+use std::path::{Path, PathBuf};
 use std::vec::Vec;
+use tokio::fs;
 
 use async_trait::async_trait;
 use clightningrpc_common::client::Client;
@@ -18,7 +20,9 @@ use coffee_lib::error;
 use coffee_lib::errors::CoffeeError;
 use coffee_lib::plugin_manager::PluginManager;
 use coffee_lib::repository::Repository;
-use coffee_lib::types::{CoffeeList, CoffeeListRemote, CoffeeRemote, CoffeeRemove};
+use coffee_lib::types::{
+    CoffeeList, CoffeeListRemote, CoffeeNurse, CoffeeRemote, CoffeeRemove, NurseStatus,
+};
 use coffee_lib::url::URL;
 use coffee_storage::file::FileStorage;
 use coffee_storage::model::repository::{Kind, Repository as RepositoryInfo};
@@ -28,27 +32,29 @@ use super::config;
 use crate::config::CoffeeConf;
 use crate::CoffeeArgs;
 
+pub type PluginName = String;
+
 #[derive(Serialize, Deserialize)]
 /// FIXME: move the list of plugin
 /// and the list of repository inside this struct.
 pub struct CoffeStorageInfo {
     pub config: config::CoffeeConf,
-    pub repositories: Vec<RepositoryInfo>,
+    pub repositories: HashMap<PluginName, RepositoryInfo>,
 }
 
 impl From<&CoffeeManager> for CoffeStorageInfo {
     fn from(value: &CoffeeManager) -> Self {
-        let mut repos = vec![];
-        // FIXME: use map instead of for each
+        let mut repos = HashMap::new();
         // FIXME: improve the down cast
-        value.repos.iter().for_each(|repo| {
+        for (name, repo) in value.repos.iter() {
             let repo = if let Some(git) = repo.as_any().downcast_ref::<Github>() {
                 RepositoryInfo::from(git)
             } else {
                 panic!("this should never happens")
             };
-            repos.push(repo);
-        });
+            repos.insert(name.to_string(), repo);
+        }
+
         CoffeStorageInfo {
             config: value.config.to_owned(),
             repositories: repos, // FIXME: found a way to downcast
@@ -58,7 +64,7 @@ impl From<&CoffeeManager> for CoffeStorageInfo {
 
 pub struct CoffeeManager {
     config: config::CoffeeConf,
-    repos: Vec<Box<dyn Repository + Send + Sync>>,
+    repos: HashMap<String, Box<dyn Repository + Send + Sync>>,
     /// Core lightning configuration managed by coffee
     coffe_cln_config: CLNConf,
     /// Core lightning configuration that include the
@@ -77,7 +83,7 @@ impl CoffeeManager {
         let mut coffee = CoffeeManager {
             config: conf.clone(),
             coffe_cln_config: CLNConf::new(conf.config_path, true),
-            repos: vec![],
+            repos: HashMap::new(),
             storage: Box::new(FileStorage::new(&conf.root_path)),
             cln_config: None,
             rpc: None,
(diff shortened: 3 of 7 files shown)


coffee_core/src/coffee.rs
```diff
@@ -34,31 +36,39 @@ use crate::CoffeeArgs;
 pub struct CoffeStorageInfo {
     pub config: config::CoffeeConf,
     pub repositories: Vec<RepositoryInfo>,
+    pub repos_names: HashSet<String>,
```
vincenzopalazzo, line 39:
```suggestion
    pub repositories: HasMap<String, RepositoryInfo>,
```

Why not this?

tareknaser360: Yes this actually makes more sense. Thanks

tareknaser360:
I changed the code to use a `HashMap<String, RepositoryInfo>` to store repository information.
I also refactored the `rm_remote` function to use the hashmap.

vincenzopalazzo (commented):
Looks good to me from a fist scan, I should test it locally tomorrow, I put it on my todo list.

P.S: at the end of may I promise you to implement a test framework for it :)

tareknaser360:
> P.S: at the end of may I promise you to implement a test framework for it :)

Great! I would love to help with that.

vincenzopalazzo (commented):
We are almost there, but this PR only addresses minor issues and does not solve the root problem. Even if we remove the `.coffee/storage.json` file (which is necessary when the JSON format changes), the refresh function still does not work, and the error persists.

My proposal is to:
- add a method to coffee called `nurse`, which will remove all the repositories in the `.coffee/repository` folder that are not in the storage. This will also clean up the plugins in the coffee configuration at the end, and it is not a trivial operation. 
- Additionally, we should run the `refresh_remote` function inside the `inventory` method, but without modifying the list of repositories. If there are any inconsistencies, we should return an error message stating that `the coffee storage is out of sync and that the user should run coffee nurse` to fix the problem.

With these changes, we should be able to solve our problem. Great job so far! Only now have I been able to get a clear idea of what needs to be done.

What do you think?

coffee_core/src/coffee.rs
```diff
@@ -288,50 +295,66 @@ impl PluginManager for CoffeeManager {
         self.storage.store(&self.storage_info()).await
     }
 
+    async fn refresh_remotes(&mut self) -> Result<(), CoffeeError> {
+        self.repos.retain(|_, repo| {
+            let repo_path = repo.url().path_string;
+            Path::new(&repo_path).exists()
```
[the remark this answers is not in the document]
vincenzopalazzo: but I do not want that the remote sync will modify this without user permission (maybe it is running coffee in the wrong directory) so we want only return the error if there is some inconsistency and tell the user how to fix it

coffee_core/src/coffee.rs
```diff
@@ -288,50 +295,66 @@ impl PluginManager for CoffeeManager {
...
+        });
+        self.storage.store(&self.storage_info()).await?;
+        Ok(())
+    }
+
     async fn add_remote(&mut self, name: &str, url: &str) -> Result<(), CoffeeError> {
+        self.refresh_remotes().await?;
```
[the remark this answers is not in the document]
tareknaser360: The problem with that is if we do, we can't call the `nurse` method because the error will be returned first and the clean up logic will not run.

coffee_core/src/coffee.rs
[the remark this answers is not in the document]
tareknaser360: I made it log a warning and return an error but not modify any files or folders.

tareknaser360:
I added a new command `nurse`
it does 2 things:
- check for repositories in `self.repos` and if one doesn't exist on disk, deletes it from storage.
- check for files and folders in `.coffee.repositories` and deletes any unrelated file (that doesn't exist in `self.repos`)

coffee_core/src/coffee.rs
```diff
@@ -288,50 +297,71 @@ impl PluginManager for CoffeeManager {
         self.storage.store(&self.storage_info()).await
     }
 
+    async fn remote_sync(&mut self) -> Result<(), CoffeeError> {
```
vincenzopalazzo, line 300:
```suggestion
    async fn remote_sync(&mut self) -> Result<(), CoffeeError> {
```

can you call this command `nurse`?

tareknaser360: The command is called nurse. This is just a function to check for the local folder that we call with every other command.

coffee_cmd/src/cmd.rs
```diff
@@ -53,6 +53,9 @@ pub enum CoffeeCommand {
     /// show the README file of the plugin
     #[clap(arg_required_else_help = true)]
     Show { plugin: String },
+    /// clean up remote repositories storage information
+    #[clap(arg_required_else_help = false)]
```
vincenzopalazzo, line 57:
```suggestion
    #[clap(arg_required_else_help = true)]
```

coffee_cmd/src/main.rs
```diff
@@ -98,6 +98,17 @@ async fn main() -> Result<(), CoffeeError> {
             }
             Err(err) => Err(err),
         },
+        CoffeeCommand::Nurse {} => {
+            let mut spinner = term::spinner(&format!("Refreshing storage"));
```
vincenzopalazzo, line 102:
```suggestion
            let mut spinner = term::spinner(&format!("Trying restoring from a corrupted status!"));
```

We should also check if there is a corrupted status, if not we return a message that there is no need to run the command

coffee_core/src/coffee.rs
```diff
@@ -257,9 +264,10 @@ impl PluginManager for CoffeeManager {
     }
 
     async fn list(&mut self, remotes: bool) -> Result<CoffeeList, CoffeeError> {
+        self.remote_sync().await?;
```
vincenzopalazzo, line 267: why not put this method inside the `inventory` method of the plugin manager?

coffee_core/src/coffee.rs
```diff
@@ -288,50 +297,71 @@ impl PluginManager for CoffeeManager {
...
         repo.init().await?;
-        self.repos.push(Box::new(repo));
+        self.repos.insert(repo.name(), Box::new(repo));
         log::debug!("remote added: {} {}", name, &url.url_string);
         self.storage.store(&self.storage_info()).await?;
         Ok(())
     }
 
     async fn rm_remote(&mut self, name: &str) -> Result<(), CoffeeError> {
+        self.remote_sync().await?;
```
vincenzopalazzo, line 330: yeah I think we need to find another place where to call this method

coffee_core/src/coffee.rs
```diff
@@ -358,6 +389,34 @@ impl PluginManager for CoffeeManager {
...
+            let entry = entry?;
+            let entry_name = entry.file_name().to_string_lossy().to_string();
+            if !self.repos.contains_key(&entry_name) {
+                if entry.file_type()?.is_dir() {
+                    fs::remove_dir_all(entry.path())?;
+                    log::debug!("folder removed: {}", entry_name);
+                } else if entry.file_type()?.is_file() {
+                    fs::remove_file(entry.path())?;
+                    log::debug!("file removed: {}", entry_name);
+                }
```
vincenzopalazzo, line 405: add a comment here that if there is a missing file inside the repository we considered the repository corrupted, it is not trivial for the people that will go to read the  code

coffee_core/src/lib.rs
```diff
@@ -14,6 +14,7 @@ pub enum CoffeeOperation {
     /// Setup(core lightning root path)
     Setup(String),
     Show(String),
+    Nurse(),
```
vincenzopalazzo, line 17:
```suggestion
    Nurse,
```

this is also valid rust right?

coffee_cmd/src/cmd.rs
tareknaser360: Should coffee nurse command take any arguments?

vincenzopalazzo (changes requested):
Some comments but the code looks really good!

I like that you put the PR when it is ready to be review, thanks

coffee_core/src/coffee.rs
tareknaser360: But then it will be called with coffee nurse itself and the error will be returned before the code runs

coffee_core/src/lib.rs
tareknaser360: Yes it is.

coffee_cmd/src/main.rs
tareknaser360: I modified the nurse function to return a json object with the status of the file and print a message to the user accordingly.

coffee_core/src/coffee.rs
vincenzopalazzo: Right, there is no way to check what kind of command the inventory is running?

tareknaser360: I searched and there was a way using an extra Boolean but this is cleaner in my opinion.

vincenzopalazzo: lets keep this and think about a possible solution in the mean while

coffee_cmd/src/cmd.rs
vincenzopalazzo: Ah now I got it, no your are right!

coffee_cmd/src/main.rs
```diff
@@ -98,6 +98,23 @@ async fn main() -> Result<(), CoffeeError> {
             }
             Err(err) => Err(err),
         },
+        CoffeeCommand::Nurse {} => {
+            let mut spinner = term::spinner(&format!("Trying restoring from a corrupted status!"));
+            match coffee.nurse().await {
+                Ok(val) => {
+                    let status = val["nurse"].as_str().unwrap();
+                    if status == "corrupt" {
+                        spinner.message("Storage refreshed");
```
vincenzopalazzo, line 107: Please make a model and not return a plan json here

vincenzopalazzo (changes requested):

coffee_cmd/src/main.rs
tareknaser360: Sure. Just pushed the code with the changes.

coffee_cmd/src/main.rs
```diff
@@ -98,6 +98,23 @@ async fn main() -> Result<(), CoffeeError> {
             }
             Err(err) => Err(err),
         },
+        CoffeeCommand::Nurse {} => {
+            let mut spinner = term::spinner(&format!("Trying restoring from a corrupted status!"));
+            match coffee.nurse().await {
+                Ok(val) => {
+                    let status = val.status;
+                    if status == "corrupt" {
```
vincenzopalazzo, line 106:
```suggestion
                    let status = val.status;
                    if status == "corrupt" {
```

It is fine, but the Rust way of doing this (also the functional way to do this, not just in Rust) is to define a type and match on this type.

So let's imagine having an enum like the following one in our model

```rust
pub enum NurseStatus {
      Currupted,
     Sane,
}
```

You than you do the match like 

```rust
match val.status {
    NurseStatus::Corrupted => {},
    NurseStatus::Sane => {},
}
```

This improves the code a lot

coffee_core/src/coffee.rs
```diff
@@ -288,50 +297,71 @@ impl PluginManager for CoffeeManager {
...
+            Some(repo) => {
+                let remote_repo = repo.list().await?;
+                let repo_path = repo.url().path_string;
+                let plugins = self.config.plugins.clone();
+                for plugin in &remote_repo {
+                    if let Some(ind) = plugins
+                        .iter()
+                        .position(|elem| elem.name() == *plugin.name())
+                    {
+                        let plugin_name = &plugins[ind].name()[..];
```
vincenzopalazzo, line 342:
```suggestion
                        let plugin_name = &plugins[ind].name().clone()
```

Do not work in this way?

vincenzopalazzo (changes requested):

coffee_cmd/src/main.rs
tareknaser360: Yes this is much cleaner

coffee_core/src/coffee.rs
tareknaser360: Yes this works

tareknaser360:
0b8daeb solves #123

coffee_core/src/coffee.rs
```diff
@@ -395,13 +394,15 @@ impl PluginManager for CoffeeManager {
     }
 
     async fn nurse(&mut self) -> Result<CoffeeNurse, CoffeeError> {
-        let repos_path = PathBuf::from(format!("{}/repositories/", self.config.root_path));
         let mut status = NurseStatus::Sane;
+
+        // delete any unrelated files or folders in `repositories` folder
+        let repos_path = PathBuf::from(format!("{}/repositories/", self.config.root_path));
         for entry in fs::read_dir(repos_path)? {
```
vincenzopalazzo, line 401: we should use tokio as `fs` otherwise you break the async workflow

vincenzopalazzo (changes requested):
ok we are super close, just a minor change in the API that you are using

tareknaser360:
This is now rebased on top of master.

Issue #131, opened 2023-05-11
allow the user to choose the branch/release of the plugin repository


vincenzopalazzo:
### Discussed in https://github.com/coffee-tools/coffee/discussions/130

<div type='discussions-op-text'>

<sup>Originally posted by **tareknaser360** May 10, 2023</sup>
Based on our earlier discussion, it was suggested to incorporate a new feature into upgrade command #43. 

Specifically, the proposed idea involves the addition of an option to select a specific branch or release of a remote repository, which can be specified by passing a flag. An example of the syntax for using this feature would be
`coffee upgrade folgore --release 2.1`

*It is important to note that this proposal assumes the presence of only one plugin in each remote repository.*

I think such feature would improve the overall user experience. </div>

tareknaser360:
A note that if a user updates the local repository to be a specific branch, we should keep the branch name stored so that if `coffee upgrade repo` was run again, the changes from that branch are pulled.

vincenzopalazzo:
yeah I am start to think that we need a more powerful persistence, so I think I need to finish my implementation of https://github.com/vincenzopalazzo/nosql-db

vincenzopalazzo:
@tareknaser we can open a issue for doing the same thing but for coffee install --branch <name>? 

This should be also easy right?

tareknaser:
> This should be also easy right?

Yes 
this is straightforward

PR #132, opened 2023-05-14
feat(coffee): implement upgrade command

### LINKED ISSUE #43: Support upgrade command to upgrade a plugin or all the plugin installed
# Background

```
cargo run -- setup <core lightning root path>
```

Currently, there is no way to upgrade a plugin and rerun the new installation, so this task required to you to implement this command

# Goal

Be able to run the following command  

Upgrade all the command that is installed by coffee, the output of this command, for now, could be just a list of JSON object

```
cargo run -- upgrade --all
```

Upgrade a particular plugin installed inside the system,  the output of this command, for now, could be just a list of JSON object

```
cargo run -- upgrade <name plugin>
```

### Additional Points

A command flag that allow you to pull the update code from the repository, but not install the plugin instead shows the list of new commits message to the user.

```
cargo run -- upgrade --show <plugin name>
```

tareknaser360:
Fixes #43 

-  The code implements the `upgrade` command
- Users can choose to upgrade a single repository or all repositories
- To upgrade a single repository, the user runs `coffee upgrade <repo_name>`
- To upgrade all repositories, the user runs `coffee upgrade --all`
- It is not possible to use the `all` flag with a specific repository name
- The `upgrade` function returns a `CoffeeUpgrade` object indicating whether the local clone is up to date or not
- A new field called `branch` has been added to the repository structure
- This field accounts for the situation where the main branch is referred to as `master`
- The `branch` field will facilitate future work on issue #131.

### DIFF
--- a/coffee_cmd/src/main.rs
+++ b/coffee_cmd/src/main.rs
@@ -7,7 +7,7 @@ use radicle_term as term;
 use coffee_core::coffee::CoffeeManager;
 use coffee_lib::errors::CoffeeError;
 use coffee_lib::plugin_manager::PluginManager;
-use coffee_lib::types::NurseStatus;
+use coffee_lib::types::{NurseStatus, UpgradeStatus};
 
 use crate::cmd::CoffeeArgs;
 use crate::cmd::CoffeeCommand;
@@ -56,7 +56,23 @@ async fn main() -> Result<(), CoffeeError> {
             let remotes = coffee.list().await;
             coffee_term::show_list(remotes)
         }
-        CoffeeCommand::Upgrade => coffee.upgrade(&[""]).await,
+        CoffeeCommand::Upgrade { repo } => {
+            match coffee.upgrade(&repo).await {
+                Ok(res) => match res.status {
+                    UpgradeStatus::UpToDate => {
+                        term::info!("Remote repository `{}` is already up to date!", res.repo)
+                    }
+                    UpgradeStatus::Updated => {
+                        term::success!(
+                            "Remote repository `{}` was successfully upgraded!",
+                            res.repo
+                        )
+                    }
+                },
+                Err(err) => return Err(err),
+            }
+            Ok(())
+        }
         CoffeeCommand::Remote { action } => match action {
             RemoteAction::Add { name, url } => {
                 let mut spinner = term::spinner(format!("Fetch remote from {url}"));
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -21,7 +21,8 @@ use coffee_lib::errors::CoffeeError;
 use coffee_lib::plugin_manager::PluginManager;
 use coffee_lib::repository::Repository;
 use coffee_lib::types::{
-    CoffeeList, CoffeeListRemote, CoffeeNurse, CoffeeRemote, CoffeeRemove, NurseStatus,
+    CoffeeList, CoffeeListRemote, CoffeeNurse, CoffeeRemote, CoffeeRemove, CoffeeUpgrade,
+    NurseStatus,
 };
 use coffee_lib::url::URL;
 use coffee_storage::file::FileStorage;
@@ -272,11 +273,22 @@ impl PluginManager for CoffeeManager {
         })
     }
 
-    async fn upgrade(&mut self, _: &[&str]) -> Result<(), CoffeeError> {
+    async fn upgrade(&mut self, repo: &str) -> Result<CoffeeUpgrade, CoffeeError> {
         self.remote_sync().await?;
-        // FIXME: Fix debug message with the list of plugins to be upgraded
-        log::debug!("upgrading plugins");
-        Ok(())
+
+        let repository = self
+            .repos
+            .get(repo)
+            .ok_or_else(|| error!("Repository with name: `{}` not found", repo))?;
+
+        let status = repository.upgrade(&self.config.plugins).await?;
+        for plugins in status.plugins_effected.iter() {
+            self.remove(plugins).await?;
+            // FIXME: pass the verbose flag to the upgrade command
+            self.install(plugins, false, false).await?;
+        }
+        self.storage.store(&self.storage_info()).await?;
+        Ok(status)
     }
 
     async fn setup(&mut self, cln_dir: &str) -> Result<(), CoffeeError> {
--- a/coffee_github/src/repository.rs
+++ b/coffee_github/src/repository.rs
@@ -1,6 +1,7 @@
 use std::any::Any;
 
 use async_trait::async_trait;
+use coffee_lib::types::CoffeeUpgrade;
 use git2;
 use log::debug;
 use tokio::fs::File;
@@ -20,6 +21,7 @@ use coffee_storage::model::repository::Kind;
 use coffee_storage::model::repository::Repository as StorageRepository;
 
 use crate::utils::clone_recursive_fix;
+use crate::utils::fast_forward;
 
 pub struct Github {
     /// the url of the repository to be able
@@ -31,6 +33,9 @@ pub struct Github {
     /// all the plugin that are listed inside the
     /// repository
     plugins: Vec<Plugin>,
+    /// the name of the branch to be able to
+    /// pull the changes from the correct branch
+    branch: String,
 }
 
 // FIXME: move this inside a utils dir craters
@@ -51,6 +56,7 @@ impl Github {
             name: name.to_owned(),
             url: url.clone(),
             plugins: vec![],
+            branch: "".to_owned(),
         }
     }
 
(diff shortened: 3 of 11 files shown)


tareknaser360:
For now this code runs only for upgrading one repository
`coffee upgrade folgore`
the feedback shows the user whether the local clone is up to date or not


There are three mutable borrow errors that need to be resolved

vincenzopalazzo (commented):
Concept ACK,

I looked at the code and seems solid, but I would like to spend some more time thinking about the current design.

For now good job

tareknaser360 (commented):
By the way there are some errors that I couldn't solve

coffee_core/src/coffee.rs
```diff
@@ -272,11 +273,77 @@ impl PluginManager for CoffeeManager {
...
+    async fn upgrade(&mut self, repo: &str, all: bool) -> Result<CoffeeUpgrade, CoffeeError> {
         self.remote_sync().await?;
-        // FIXME: Fix debug message with the list of plugins to be upgraded
-        log::debug!("upgrading plugins");
-        Ok(())
+        let mut total_status: Vec<CoffeeUpgradeOne> = vec![];
+        if all {
+            // for repository in self.repos.keys() {
+            //     match self.upgrade_single(repository).await {
+            //         Ok(val) => total_status.push(val),
```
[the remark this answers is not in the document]
vincenzopalazzo: Feedback that I had from a design point of view, it is possible to write the `async fn upgrade_single(&mut self, repo: &str) -> Result<CoffeeUpgradeOne, CoffeeError> {` inside the single repository?

This should remove a borrowing problem, but also this is the right design because the repository knows how to upgrade itself.

The only think that I am not sure is we the repository to upgrade itself need to know somethings from outside the repository.

Sorry I did not look at the code yet

vincenzopalazzo (commented):
To solve this i need to return and run the code but these some inputs

vincenzopalazzo (changes requested):
For now, my focus has been on fixing the compiler and working on a better separation of concerns. However, I haven't had a chance to thoroughly examine the solution itself.

Please cherry-pick commit https://github.com/coffee-tools/coffee/commit/d11212567f1801ceb2d1447305af844e2c417614. The motivation behind this change is explained within the commit itself. :)

Let me know if you need more information to understand it.

tareknaser360:
> Let me know if you need more information to understand it.

I went through the commit and it made sense to me.

I made slight changes to fix some clippy warnings. 
Let me know if the solution itself is what we wanted.

coffee_core/src/coffee.rs
```diff
@@ -272,11 +273,77 @@ impl PluginManager for CoffeeManager {
...
+                    }
+                    UpgradeStatus::Updated => {
+                        // uninstall the plugins
+                        // for plugin in &installed_plugins {
+                        //     self.remove(&plugin).await?;
+                        // }
+                        // reinstall the plugins
+                        // for plugin in &installed_plugins {
+                        //     self.install(&plugin, false, false).await?;
+                        // }
```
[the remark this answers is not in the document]
tareknaser360: But then if a plugin changed the name (path) of the executable in an upgrade/branch, we would have a problem because the older version of the exec_path would still exist in the configuration.

coffee_cmd/src/main.rs
```diff
@@ -56,7 +56,29 @@ async fn main() -> Result<(), CoffeeError> {
             let remotes = coffee.list().await;
             coffee_term::show_list(remotes)
         }
-        CoffeeCommand::Upgrade => coffee.upgrade(&[""]).await,
+        CoffeeCommand::Upgrade { repo, all } => {
+            let result = match repo {
+                Some(repo) => coffee.upgrade(&repo, all).await,
+                None => coffee.upgrade("", all).await,
```
vincenzopalazzo, line 62:
```suggestion
                None => coffee.upgrade("", all).await,
```

this is the meaning of all? why we need to pass the `""` empty string? we can use the option here

vincenzopalazzo: you can also remove the `all` because if the `repo` is none at this point the upgrade command will upgrade all the repository

coffee_core/src/coffee.rs
```diff
@@ -272,11 +273,31 @@ impl PluginManager for CoffeeManager {
         })
     }
 
-    async fn upgrade(&mut self, _: &[&str]) -> Result<(), CoffeeError> {
+    async fn upgrade(&mut self, repo: &str, all: bool) -> Result<CoffeeUpgrade, CoffeeError> {
```
vincenzopalazzo, line 276:
```suggestion
    async fn upgrade(&mut self, repo: Option<String>) -> Result<CoffeeUpgrade, CoffeeError> {
```

Maybe this is a good signature?

coffee_core/src/coffee.rs
```diff
@@ -272,11 +273,31 @@ impl PluginManager for CoffeeManager {
...
         self.remote_sync().await?;
-        // FIXME: Fix debug message with the list of plugins to be upgraded
-        log::debug!("upgrading plugins");
-        Ok(())
+        let mut total_status: Vec<CoffeeRepositoryUpgrade> = vec![];
+        if all {
```
vincenzopalazzo, line 279:
```suggestion
        if repo.is_none() {
```

and then you can do somethings like that

coffee_core/src/coffee.rs
```diff
@@ -272,11 +273,31 @@ impl PluginManager for CoffeeManager {
...
+            for repo in self.repos.values() {
+                let result = repo.upgrade(&self.config.plugins).await?;
+                total_status.push(result);
+            }
+        } else {
+            let repository = self
+                .repos
+                .get(repo)
+                .ok_or(error!("repository with name: {repo} not found"))?;
+            let status = repository.upgrade(&self.config.plugins).await?;
```
vincenzopalazzo, line 289: Looks like that there is a log somewhere

By running the following command

```
➜  coffee git:(pr132-review) ✗ cargo run --bin coffee -- --data-dir /home/vincent/clightning4j --network testnet upgrade lightningd
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/coffee --data-dir /home/vincent/clightning4j --network testnet upgrade lightningd`
[2023-05-18T02:20:13Z ERROR coffee_lib::errors] ERROR #1: repository with name: lightningd not found
Remote repository lightningd is already up to date!
➜  coffee git:(pr132-review) ✗ 

```

Why we have the `ERROR coffee_lib::errors] ERROR #1:`?

coffee_core/src/coffee.rs
```diff
@@ -272,11 +273,31 @@ impl PluginManager for CoffeeManager {
...
+        if all {
+            for repo in self.repos.values() {
+                let result = repo.upgrade(&self.config.plugins).await?;
+                total_status.push(result);
+            }
+        } else {
+            let repository = self
+                .repos
+                .get(repo)
+                .ok_or(error!("repository with name: {repo} not found"))?;
```
vincenzopalazzo, line 288:
```suggestion
                .ok_or(error!("repository with name: `{repo}` not found"))?;
```

coffee_cmd/src/main.rs
```diff
@@ -56,7 +56,29 @@ async fn main() -> Result<(), CoffeeError> {
...
+                            UpgradeStatus::Updated => term::success!(
+                                "Remote repository {} was successfully upgraded!",
+                                res.repo
+                            ),
+                        }
+                    }
+                }
+                Err(err) => term::error(format!("{err}")),
+            }
+            Ok(())
```
vincenzopalazzo, line 80: Mh, this upgrade is all creating problem, if one fails we should abort the upgrade, otherwise, we leave the job not completely done.

For now we do not allow upgrading a single repository by name only? what do you think?

vincenzopalazzo (changes requested):
I had some questions and also I think that for now, we should keep the command simple.

I think that for now upgrading a single repository at a time is good and in the mean while we try to think of a possible solution to this problem.

coffee_core/src/coffee.rs
tareknaser360: I used `ok_or_else` instead of `ok_or` to fix this
It is invoked lazily, meaning the error message expression is only evaluated if the Option is None.

coffee_core/src/coffee.rs
tareknaser360: Yes this makes more sense.
I will use this approach when I work on adding support for `all` flag

coffee_cmd/src/main.rs
```diff
@@ -56,7 +56,21 @@ async fn main() -> Result<(), CoffeeError> {
             let remotes = coffee.list().await;
             coffee_term::show_list(remotes)
         }
-        CoffeeCommand::Upgrade => coffee.upgrade(&[""]).await,
+        CoffeeCommand::Upgrade { repo } => {
+            let result = coffee.upgrade(&repo).await;
+            match result {
```
vincenzopalazzo, line 61:
```suggestion
            match coffee.upgrade(&repo).await {
```

this should work

coffee_cmd/src/main.rs
```diff
@@ -56,7 +56,21 @@ async fn main() -> Result<(), CoffeeError> {
...
+                Ok(res) => match res.status {
+                    UpgradeStatus::UpToDate => {
+                        term::info!("Remote repository {} is already up to date!", res.repo)
```
vincenzopalazzo, line 64:
```suggestion
                        term::info!("Remote repository `{}` is already up to date!", res.repo)
```

coffee_cmd/src/main.rs
```diff
@@ -56,7 +56,21 @@ async fn main() -> Result<(), CoffeeError> {
...
+                    }
+                    UpgradeStatus::Updated => {
+                        term::success!("Remote repository {} was successfully upgraded!", res.repo)
```
vincenzopalazzo, line 67:
```suggestion
                        term::success!("Remote repository `{}` was successfully upgraded!", res.repo)
```

coffee_cmd/src/main.rs
```diff
@@ -56,7 +56,21 @@ async fn main() -> Result<(), CoffeeError> {
...
+                    }
+                },
+                Err(err) => term::error(format!("{err}")),
```
vincenzopalazzo, line 70:
```suggestion
                Err(err) => Err(err),
```

We are printing the error already outside the match, right?

coffee_github/src/repository.rs
```diff
@@ -197,6 +203,11 @@ impl Repository for Github {
         let res = git2::Repository::clone(&self.url.url_string, &self.url.path_string);
         match res {
             Ok(repo) => {
+                self.branch = if repo.find_branch("main", git2::BranchType::Local).is_ok() {
```
vincenzopalazzo, line 206:
```suggestion
                self.branch = if repo.find_branch("master", git2::BranchType::Local).is_ok() {
```

vincenzopalazzo: lets take  by default the master branch that it is more common

coffee_lib/src/types/mod.rs
```diff
@@ -35,3 +35,19 @@ pub enum NurseStatus {
...
+    UpToDate,
+    Updated,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct CoffeeUpgrade {
+    pub repo: String,
+    pub status: UpgradeStatus,
+    /// If the status of the repository is
+    /// alternate we return the list of plugin
```
vincenzopalazzo, line 50:
```suggestion
    /// alterate we return the list of plugin
```

vincenzopalazzo (changes requested):
Just a minors nits

coffee_cmd/src/main.rs
tareknaser360: Yes

vincenzopalazzo (approved):
We need a way to test it, but the code looks like solid!

Thanks, and really good work!

Issue #128, opened 2023-05-10
Nurse Command Recovery Strategy Cases and Handlers


tareknaser:
|**Case**|**Handler**|
|---|---|
|1. A `git` repository exists in `coffee` configuration but absent from local storage|- &#9744; Attempt to clone the repository using the `git` HEAD reference stored in the configuration.<br/>- &#9744; If the cloning process fails, remove the repository entry from the configuration.|
|2. A repository is present in local storage (`./coffee/repositories`) but is not documented in `coffee` configuration|- &#9744; Make an attempt to index the repository, verifying if it is a `git` repository, and subsequently add it to the `coffee` configuration.<br/>- &#9744; If indexing fails, the repository will be removed from the `repositories` folder.|
|3. A `git` repository is found to be corrupt, typically due to a missing file or other issues|- &#9744; Recover using the `git` HEAD.|

**Note: Before proceeding with any of these actions, we have to ensure that the user has confirmed**

vincenzopalazzo:
We should also try to recover the state of the storage is the git repository has no changes

vincenzopalazzo:
Ok I noted that if the `.coffee/repository` do not exist the call `self.remote_sync` fails.

So this is wrong for some commands like add remote, `self.remote_sync` fails when there is no repository directory
we will be never able to add a new remote.

This feature looks untested. Are you testing your PR on your system?

tareknaser360:
Yes I think I missed this corner case
I tested the function locally but this issue only happens when the user sets up coffee the first time

vincenzopalazzo:
>I tested the function locally but this issue only happens when the user sets up coffee the first time

or you use the --data-dir, so the test that we need to run each time is create a new directory and make sure that we do not introduce regression

PR #133, opened 2023-05-18
fix(core): allow dirty state for add remote

vincenzopalazzo:
See https://github.com/coffee-tools/coffee/issues/128#issuecomment-1552318097 and commit for details

### DIFF
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -315,7 +315,11 @@ impl PluginManager for CoffeeManager {
     }
 
     async fn add_remote(&mut self, name: &str, url: &str) -> Result<(), CoffeeError> {
-        self.remote_sync().await?;
+        // FIXME: we should allow some error here like
+        // for the add remote command the no found error for the `repository`
+        // directory is fine.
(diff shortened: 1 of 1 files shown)


coffee_core/src/coffee.rs
```diff
@@ -315,7 +315,11 @@ impl PluginManager for CoffeeManager {
     }
 
     async fn add_remote(&mut self, name: &str, url: &str) -> Result<(), CoffeeError> {
-        self.remote_sync().await?;
+        // FIXME: we should allow some error here like
```
tareknaser360, line 318: Maybe we can call the function only if the directory exists?
```
if Path::new(&format!("{}/repositories/", self.config.root_path)).exists() {
            self.remote_sync().await?;
}
```

tareknaser360 (changes requested):

coffee_core/src/coffee.rs
vincenzopalazzo: No this should be a generic step; I need to think a little bit more.

The nurse command is broken, and we should improve it and not spread this hack across the code base imho.

PR #134, opened 2023-05-21
feat(httpd): implement list call

tareknaser360:
- Added a new function coffee_list to handle the `/list` endpoint.
- The function takes the data parameter of type `web::Data<AppState>` to access the shared state.
- The function doesn't take any body

This is the response from the server for this call
image

### DIFF
--- a/coffee_httpd/Cargo.toml
+++ b/coffee_httpd/Cargo.toml
@@ -6,7 +6,10 @@ edition = "2021"
 [dependencies]
 actix-web = "4"
 clap = "4.1.11"
+tokio = { version = "1.22.0", features = ["sync"] }
 coffee_core = { path = "../coffee_core" }
+coffee_lib = { path = "../coffee_lib" }
 paperclip = { version = "0.8.0", features = ["actix4"] }
 log = "0.4.17"
 env_logger = "0.9.3"
+serde_json = "1"
--- a/coffee_httpd/src/httpd/server.rs
+++ b/coffee_httpd/src/httpd/server.rs
@@ -9,8 +9,12 @@
 use std::collections::HashMap;
 use std::net::ToSocketAddrs;
 use std::sync::Arc;
+use tokio::sync::Mutex;
+
+use serde_json::Value;
 
 use coffee_core::coffee::CoffeeManager;
+use coffee_lib::plugin_manager::PluginManager;
 
 use actix_web::{App, HttpResponse};
 use actix_web::{Error, HttpServer};
@@ -29,7 +33,7 @@ use paperclip::actix::{
 // This struct represents state
 struct AppState {
     #[allow(dead_code)]
-    coffee: Arc<CoffeeManager>,
+    coffee: Arc<Mutex<CoffeeManager>>,
 }
 
 /// entry point of the httd to allow
@@ -38,7 +42,7 @@ pub async fn run_httpd<T: ToSocketAddrs>(
     coffee: CoffeeManager,
     host: T,
 ) -> Result<(), std::io::Error> {
-    let rc = Arc::new(coffee);
+    let rc = Arc::new(Mutex::new(coffee));
     HttpServer::new(move || {
         let state = AppState { coffee: rc.clone() };
         App::new()
@@ -46,6 +50,7 @@ pub async fn run_httpd<T: ToSocketAddrs>(
             .wrap_api()
             .service(swagger_api)
             .service(coffee_help)
+            .service(coffee_list)
             .with_json_spec_at("/api/v1")
             .build()
     })
@@ -62,6 +67,24 @@ async fn coffee_help(
 ) -> Result<Json<HashMap<String, String>>, Error> {
     // FIXME: the json need to be a struct
     Ok(body)
+}
+
+#[api_v2_operation]
+#[get("/list")]
+async fn coffee_list(data: web::Data<AppState>) -> Result<Json<Value>, Error> {
+    let mut coffee = data.coffee.lock().await;
+    let result = coffee.list().await;
+    match result {
+        Ok(coffee_list) => {
+            let val = serde_json::to_value(coffee_list).map_err(|err| {
+                actix_web::error::ErrorInternalServerError(format!("coffee list error: {err}"))
+            })?;
+            Ok(Json(val))
+        }
+        Err(err) => Err(actix_web::error::ErrorInternalServerError(format!(
+            "coffee list error: {err}"
+        ))),
+    }
 }
 
 // this is just an hack to support swagger UI with https://paperclip-rs.github.io/paperclip/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -490,9 +490,12 @@ dependencies = [
  "actix-web",
  "clap",
  "coffee_core",
+ "coffee_lib",
  "env_logger",
  "log",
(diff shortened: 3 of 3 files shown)


coffee_httpd/Cargo.toml
```diff
@@ -6,7 +6,9 @@ edition = "2021"
 [dependencies]
 actix-web = "4"
 clap = "4.1.11"
+tokio = { version = "1.22.0", features = ["full"] }
```
vincenzopalazzo, line 9:
```suggestion
tokio = { version = "1.22.0", features = ["full"] }
```

we really need the full feature?

coffee_httpd/src/httpd/server.rs
```diff
@@ -64,6 +68,39 @@ async fn coffee_help(
     Ok(body)
 }
 
+#[api_v2_operation]
+#[get("/list")]
+async fn coffee_list(
+    data: web::Data<AppState>,
+) -> Result<Json<HashMap<String, Vec<HashMap<String, String>>>>, Error> {
+    let mut coffee = data.coffee.lock().await;
+    let result = coffee.list().await;
```
vincenzopalazzo, line 77:
```suggestion
    let result = coffee.list().await;
```

Why the coffee list is not returning directly a model that we can you inside the JSON response? as 

```rust
json!({ "plugin": result })
```

vincenzopalazzo (changes requested):

coffee_httpd/src/httpd/server.rs
tareknaser360: We can but there will be additional information shown to the user.

coffee_httpd/src/httpd/server.rs
```diff
@@ -64,6 +69,22 @@ async fn coffee_help(
...
 
+#[api_v2_operation]
+#[get("/list")]
+async fn coffee_list(data: web::Data<AppState>) -> Result<Json<Value>, Error> {
+    let mut coffee = data.coffee.lock().await;
+    let result = coffee.list().await;
+    match result {
+        Ok(coffee_list) => {
+            let plugins: Vec<Plugin> = coffee_list.plugins;
+            Ok(Json(json!({ "Plugins": plugins })))
```
vincenzopalazzo, line 80:
```suggestion
            Ok(Json(json!({ "plugins": coffee_list.plugins })))
```

coffee_httpd/src/httpd/server.rs
```diff
@@ -64,6 +69,22 @@ async fn coffee_help(
...
 }
 
+#[api_v2_operation]
+#[get("/list")]
+async fn coffee_list(data: web::Data<AppState>) -> Result<Json<Value>, Error> {
+    let mut coffee = data.coffee.lock().await;
+    let result = coffee.list().await;
+    match result {
+        Ok(coffee_list) => {
+            let plugins: Vec<Plugin> = coffee_list.plugins;
```
vincenzopalazzo, line 79:
```suggestion
```

coffee_httpd/Cargo.toml
```diff
@@ -6,7 +6,10 @@ edition = "2021"
 [dependencies]
 actix-web = "4"
 clap = "4.1.11"
+tokio = "1.22.0"
```
vincenzopalazzo, line 9:
```suggestion
tokio = "1.22.0"
```

I think this is still wrong, you should brink inside only the feature of tokio that brink the `sync` package, take a look to the tokio rust docs, it is a well documented package

vincenzopalazzo (changes requested):

vincenzopalazzo:
>Now the return value looks like this.


I preferer that you put the json in a text box, the screenshot is quite unreadble

```json
{ 
    "plugins": {}
}
```

coffee_httpd/src/httpd/server.rs
vincenzopalazzo: The Rest response and the core response should be the same, so what additional response are you returning to the user?

coffee_httpd/src/httpd/server.rs
vincenzopalazzo: Ah no, this is wrong, and your code too.

There is no need to use the `json` macro; you can use the value directly
`Json(coffee_list)` or if the Json is not so smart, you can do `Json(serde_json::to_value(coffee_list))`

The `coffee_list` is already the correct format that we want

coffee_httpd/Cargo.toml
tareknaser360: I only included the `sync` module from tokio

tareknaser360:
This is the return value. For the 2 cases whether one of the plugins includes coffee conf file or not
```
{
   "plugins": [
      {
         "conf": {
            "plugin": {
               "dependencies": null,
               "deprecated": null,
               "install": "make coffee\n",
               "lang": "rust",
               "main": "target/release/folgore_plugin",
               "name": "folgore",
               "version": "0.0.1"
            }
         },
         "exec_path": "/Users/tareknasser/.coffee/repositories/fol/target/release/folgore_plugin",
         "lang": "Rust",
         "name": "folgore",
         "root_path": "/Users/tareknasser/.coffee/repositories/fol"
      },
      {
         "conf": null,
         "exec_path": "/Users/tareknasser/.coffee/repositories/lightningd/summary/summary.py",
         "lang": "PyPip",
         "name": "summary",
         "root_path": "/Users/tareknasser/.coffee/repositories/lightningd/summary"
      },
      {
         "conf": null,
         "exec_path": "/Users/tareknasser/.coffee/repositories/lightningd/helpme/helpme.py",
         "lang": "PyPip",
         "name": "helpme",
         "root_path": "/Users/tareknasser/.coffee/repositories/lightningd/helpme"
      }
   ]
}
```

coffee_httpd/src/httpd/server.rs
tareknaser360: In this case our return value would be of type `Json<CoffeeList>` which causes errors
We would have to implement the `TypedData` and `Apiv2Schema` traits for `CoffeeList` type.

tareknaser360: We can do this
```
let result = coffee.list().await;
match result {
      Ok(coffee_list) => match serde_json::to_value(coffee_list) {
          Ok(val) => Ok(Json(val)),
          Err(_) => Err(actix_web::error::ErrorInternalServerError(
              "coffee list error",
          )),
      },
      Err(_) => Err(actix_web::error::ErrorInternalServerError(
          "coffee list error",
      )),
```
and the return value remain of type `Json<Value>`
Is this better than the current changes?

vincenzopalazzo: yes it is better

vincenzopalazzo: but please do not give the coffee error

```
let val = serde_json::to_value(coffee_list)
            .map_err(|err| actix_web::error::ErrorInternalServerError(format!("coffee list error: {err}")))?;

```

if this do not work, implement the `std::fmt::Display` on the `CoffeeError`

tareknaser360: Works fine. I tested it as well

vincenzopalazzo (approved):
Good Job!

We should start also do make a doc page for this when we start to support all the commands :)

Issue #141, opened 2023-05-28
bad url inside the plugin struct in some cases


vincenzopalazzo:
While testing the https://github.com/coffee-tools/coffee/pull/134 we should find the cause and normalize the URL

I noted that the URL does not look right

```json 
{
  "plugins": [
    {
      "conf": null,
      "exec_path": "/tmp//.coffee/repositories/official/backup/backup.py",
      "lang": "PyPoetry",
      "name": "backup",
      "root_path": "/tmp//.coffee/repositories/official/backup"
    },
    {
      "conf": null,
      "exec_path": "/tmp//.coffee/repositories/official/summary/summary.py",
      "lang": "PyPip",
      "name": "summary",
      "root_path": "/tmp//.coffee/repositories/official/summary"
    }
  ]
}

```

there is an extra `/` in `"exec_path": "/tmp//.coffee/repositories/official/backup/backup.py"`

vincenzopalazzo:
This can be solved by https://github.com/coffee-tools/coffee/pull/142#pullrequestreview-1451027324 but need to be tested in various test cases

tareknaser360:
Is this solved?

vincenzopalazzo:
I do no know :) we should write a test for confirm

PR #142, opened 2023-05-29
fix(core): avoid double slash

tareknaser360:
Related to #141

### DIFF
--- a/coffee_core/src/config.rs
+++ b/coffee_core/src/config.rs
@@ -46,8 +46,8 @@ impl CoffeeConf {
             def_path = data_dir.to_owned();
         }
 
-        // FIXME: check for double slash
-        def_path += "/.coffee";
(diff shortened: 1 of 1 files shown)


coffee_core/src/config.rs
```diff
@@ -46,8 +46,10 @@ impl CoffeeConf {
             def_path = data_dir.to_owned();
         }
 
-        // FIXME: check for double slash
-        def_path += "/.coffee";
+        if !def_path.ends_with('/') {
+            def_path += "/";
+        }
```
vincenzopalazzo, line 51:
```suggestion
        def_path = def_path.strip_suff('/').unwrap_or(def_path);
```

This is a better way to write this in rust? or maybe it is just my preference!

vincenzopalazzo (changes requested):

Issue #145, opened 2023-06-13
Coffee do not take into count the network


vincenzopalazzo:
Coffee does not take into count the network so in this case I am not able to set the testnet network

```
root@vultr:~ # RUST_LOG=debug coffee --network testnet setup /root/.lightning
[2023-06-13T13:23:37Z INFO  coffee_core::config] creating coffee home at /root/.coffee
[2023-06-13T13:23:37Z INFO  coffee_core::coffee] looking for the cln config: /root/.lightning//bitcoin/config
[2023-06-13T13:23:37Z DEBUG coffee_core::coffee] {
        "comment # core lightning configuration generated by coffe please do not edit this": [
            "# core lightning configuration generated by coffe please do not edit this",
        ],
    }
[2023-06-13T13:23:37Z DEBUG coffee_core::coffee] cln conf CLNConf { fields: {"comment # core lightning configuration generated by coffe please do not edit this": ["# core lightning configuration generated by coffe please do not edit this"], "plugin": ["/root/.coffee/repositories/folgore-git/target/release/folgore_plugin"]}, includes: [], path: "/root/.coffee/testnet/coffee.conf", create_if_missing: true }
[2023-06-13T13:23:37Z DEBUG coffee_core::coffee] finish pligin manager inventory
[2023-06-13T13:23:37Z WARN  coffee_core::coffee] you are ovveriding the previous set up
[2023-06-13T13:23:37Z INFO  coffee_core::coffee] configure coffe in the following cln config /root/.lightning/bitcoin/config
[2023-06-13T13:23:37Z INFO  coffee_core::coffee] looking for the cln config: /root/.lightning/bitcoin/config
[2023-06-13T13:23:37Z DEBUG coffee_core::coffee] {
        "comment # core lightning configuration generated by coffe please do not edit this": [
            "# core lightning configuration generated by coffe please do not edit this",
        ],
    }
[2023-06-13T13:23:37Z INFO  coffee_core::coffee] cln configured
root@vultr:~ # coffee --help
A plugin manager for core lightning

```

vincenzopalazzo:
Should be fixed by https://github.com/coffee-tools/coffee/pull/162

PR #149, opened 2023-06-20
Refactor configuration handling to preserve user-defined setting

tareknaser360:
Addresses #145 
Coffee does not consider the network parameter when setting up the configuration. 
Currently, the configuration is being overwritten with the stored configuration, which prevents users from setting the desired network. 

By removing the line that overwrites the configuration in the inventory function, the user will be able to explicitly set the network parameter.


vincenzopalazzo (approved):
ACK  https://github.com/coffee-tools/coffee/pull/149/commits/9872143ad513ece59229b9ef6c324722550d19e8

Looks correct!

PR #152, opened 2023-06-21
fix(core): create repositories folder when configuring coffee

### LINKED ISSUE #151: Coffee doesn't give a proper feedback if the user doesn't add a remote repository first
If the user tries to run `coffee remote list` or `coffee install` for example without first adding a remote repository, coffee fails to report the problem. 
The only feedback the user is getting will be
```
✗ code: 1, msg: No such file or directory (os error 2)
```
This has to do with the implementation of the `remote_sync` function

A more user friendly feedback would be to advise the user to add a remote repository first

tareknaser360:
Fixes #151

So this is a double solution to make sure the issue is solved.
1- we create the `repositories` folder beforehand along with the `bitcoin` and `testnet` folders. 
- This by itself should solve the issue.

2- we provide a proper error message to the user if the folder is missing.

### DIFF
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -300,7 +300,9 @@ impl PluginManager for CoffeeManager {
     async fn remote_sync(&mut self) -> Result<(), CoffeeError> {
         // check if there are any unrelated files or folders in the `repositories` folder
         let repos_path = PathBuf::from(format!("{}/repositories/", self.config.root_path));
-        let mut dir_entries = fs::read_dir(repos_path).await?;
+        let mut dir_entries = fs::read_dir(repos_path)
+            .await
+            .map_err(|_| CoffeeError::new(1, "`repositories` folder wasn't found"))?;
         while let Some(entry) = dir_entries.next_entry().await? {
             let entry_name = entry.file_name().to_string_lossy().to_string();
             if !self.repos.contains_key(&entry_name) {
--- a/coffee_core/src/config.rs
+++ b/coffee_core/src/config.rs
@@ -52,6 +52,7 @@ impl CoffeeConf {
         info!("creating coffee home at {def_path}");
         check_dir_or_make_if_missing(format!("{def_path}/bitcoin")).await?;
         check_dir_or_make_if_missing(format!("{def_path}/testnet")).await?;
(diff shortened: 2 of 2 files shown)


coffee_core/src/coffee.rs
```diff
@@ -300,7 +300,9 @@ impl PluginManager for CoffeeManager {
     async fn remote_sync(&mut self) -> Result<(), CoffeeError> {
         // check if there are any unrelated files or folders in the `repositories` folder
         let repos_path = PathBuf::from(format!("{}/repositories/", self.config.root_path));
-        let mut dir_entries = fs::read_dir(repos_path).await?;
+        let mut dir_entries = fs::read_dir(repos_path)
+            .await
+            .map_err(|_| CoffeeError::new(1, "/.coffee/repositories folder wasn't found"))?;
```
vincenzopalazzo, line 305:
```suggestion
            .map_err(|_| CoffeeError::new(1, "`repositories` folder wasn't found"))?;
```

coffee_core/src/config.rs
```diff
@@ -52,6 +52,7 @@ impl CoffeeConf {
         info!("creating coffee home at {def_path}");
         check_dir_or_make_if_missing(format!("{def_path}/bitcoin")).await?;
         check_dir_or_make_if_missing(format!("{def_path}/testnet")).await?;
+        check_dir_or_make_if_missing(format!("{def_path}/repositories")).await?;
```
vincenzopalazzo, line 55: why we need this here?

vincenzopalazzo (changes requested):

coffee_core/src/config.rs
tareknaser360: If we don't create it when configuring at first, the user will get an error if he tries to use any command before adding a remote repository.
For example, if he tries to list the remotes, he will get an error saying that the `repositories` folder wasn't found.

I think it makes more sense to create it beforehand and save the error for the case that the user actually messes it up somehow by accident.

vincenzopalazzo: yes it does! thanks

could you add this explanation inside the commit body? so we know why you change this code

vincenzopalazzo (approved):
ACK https://github.com/coffee-tools/coffee/pull/152/commits/e9f109dcb2c0e9573db9257292c0f50cfc4a2139

PR #156, opened 2023-06-25
feat(test): add a new test to mimic cmd arguments

tareknaser360:
This PR adds a new test to replicate this error #145 
The test now fails but will come in handy to solve the issue

### DIFF
--- a/coffee_testing/src/lib.rs
+++ b/coffee_testing/src/lib.rs
@@ -71,14 +71,14 @@ impl coffee_core::CoffeeArgs for CoffeeTestingArgs {
 /// we need to perform integration testing for coffee.
 pub struct CoffeeTesting {
     inner: CoffeeManager,
-    root_path: TempDir,
+    root_path: String,
 }
 
 impl Drop for CoffeeTesting {
     fn drop(&mut self) {
         use std::fs;
 
-        fs::remove_dir_all(self.root_path.path()).unwrap();
+        fs::remove_dir_all(&self.root_path).unwrap();
     }
 }
 
@@ -96,11 +96,32 @@ impl CoffeeTesting {
             .map_err(|err| anyhow::anyhow!("{err}"))?;
         Ok(Self {
             inner: coffee,
-            root_path: dir,
+            root_path: dir.path().to_str().unwrap().to_owned(),
+        })
+    }
+
+    // init coffee in a tmp directory with arguments.
+    pub async fn tmp_with_args(
+        args: &CoffeeTestingArgs,
+        tempdir: &TempDir,
+    ) -> anyhow::Result<Self> {
+        log::info!("Temporary directory: {:?}", tempdir);
+
+        let coffee = CoffeeManager::new(args)
+            .await
+            .map_err(|err| anyhow::anyhow!("{err}"))?;
+
+        Ok(Self {
+            inner: coffee,
+            root_path: tempdir.path().to_str().unwrap().to_owned(),
         })
     }
 
     pub fn coffee(&mut self) -> &mut CoffeeManager {
         &mut self.inner
+    }
+
+    pub fn root_path(&self) -> &str {
+        &self.root_path
     }
 }
--- a/coffee_testing/src/lib.rs
+++ b/coffee_testing/src/lib.rs
@@ -7,6 +7,7 @@ pub mod prelude {
     pub use cln_test;
 
     pub use crate::macros::*;
+    pub use tempfile;
 }
 
 use tempfile::TempDir;
@@ -71,15 +72,7 @@ impl coffee_core::CoffeeArgs for CoffeeTestingArgs {
 /// we need to perform integration testing for coffee.
 pub struct CoffeeTesting {
     inner: CoffeeManager,
-    root_path: String,
-}
-
-impl Drop for CoffeeTesting {
-    fn drop(&mut self) {
-        use std::fs;
-
-        fs::remove_dir_all(&self.root_path).unwrap();
-    }
+    root_path: TempDir,
 }
 
 impl CoffeeTesting {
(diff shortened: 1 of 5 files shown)


tests/src/coffee_integration_tests.rs
```diff
@@ -32,6 +33,28 @@ pub async fn init_coffee_test() -> anyhow::Result<()> {
...
+        network: Some("testnet".to_string()),
+    };
+    let coffee = CoffeeManager::new(&args).await?;
+
+    let actual_network = coffee.storage_info().config.network;
+    let expected_network = "testnet".to_string();
+
+    assert_eq!(
+        actual_network, expected_network,
+        "Network is wrong. Actual: '{}', Expected: '{}'",
```
vincenzopalazzo, line 52:
```suggestion
        "Network is wrong. Actual: '{actual_network}', Expected: '{expected_network}'",
```

tests/src/coffee_integration_tests.rs
```diff
@@ -32,6 +33,28 @@ pub async fn init_coffee_test() -> anyhow::Result<()> {
...
+        actual_network, expected_network
```
vincenzopalazzo, line 53:
```suggestion
```

tests/src/coffee_integration_tests.rs
```diff
@@ -32,6 +33,28 @@ pub async fn init_coffee_test() -> anyhow::Result<()> {
...
 }
 
+#[tokio::test]
+pub async fn init_coffee_test_cmd() -> anyhow::Result<()> {
+    init();
+
+    let args = MockCoffeeArgs {
+        conf: None,
+        data_dir: None,
+        network: Some("testnet".to_string()),
```
vincenzopalazzo, line 43:
```suggestion
        network: Some("regtest".to_string()),
```

Try this kind of network

vincenzopalazzo (changes requested):

coffee_testing/src/lib.rs
```diff
@@ -100,6 +101,19 @@ impl CoffeeTesting {
         })
     }
 
+    // init coffee in a tmp directory with arguments.
+    pub async fn tmp_with_args(args: &CoffeeTestingArgs) -> anyhow::Result<Self> {
+        let temp_dir_path = args.data_dir.clone();
+        let dir = Builder::new().prefix(&temp_dir_path).tempdir()?;
```
vincenzopalazzo, line 107: this is always the correct path?
```suggestion
        let dir = Builder::new().prefix(&temp_dir_path).tempdir()?;
```

vincenzopalazzo (commented):
Nice! finally we had the reproducer

coffee_testing/src/lib.rs
tareknaser360: we can take the temp path as an argument to avoid this line of code

vincenzopalazzo: why not put the tmp path directly inside the args?

tareknaser360: The args expects a string for the data directory.
what do you think of the latest commit?
I added an info debug statement to make sure they are the same

vincenzopalazzo: >The args expects a string for the data directory.

but you can convert the tmp directory to a path and then convert the path to a string no? as I did here https://github.com/coffee-tools/coffee/blob/master/coffee_testing/src/lib.rs#L90

coffee_testing/src/lib.rs
```diff
@@ -100,6 +100,21 @@ impl CoffeeTesting {
         })
     }
 
+    // init coffee in a tmp directory with arguments.
+    pub async fn tmp_with_args(args: &CoffeeTestingArgs) -> anyhow::Result<Self> {
+        let tempdir = tempfile::tempdir_in(args.data_dir.clone())?;
+        log::info!("tempdir: {:?}", tempdir);
```
vincenzopalazzo, line 106: Why do not pass the dir as parameter? No one is stopping you do to it

tests/src/coffee_integration_tests.rs
```diff
@@ -32,6 +31,40 @@ pub async fn init_coffee_test() -> anyhow::Result<()> {
...
 }
 
+#[tokio::test]
+pub async fn init_coffee_test_cmd() -> anyhow::Result<()> {
+    init();
+
+    let dir = tempfile::tempdir()?;
+    let args = CoffeeTestingArgs {
+        conf: None,
+        data_dir: dir.path().to_str().unwrap().to_owned(),
```
vincenzopalazzo, line 41:
```rust
let tmp_path = dir.path().to_str().unwrap().to_owned();
```

and then use it around the code

vincenzopalazzo (changes requested):

vincenzopalazzo:
This should fix your error, and in my machine the tests is passing

```diff
diff --git a/coffee_testing/src/lib.rs b/coffee_testing/src/lib.rs
index e1dcd50..255b988 100644
--- a/coffee_testing/src/lib.rs
+++ b/coffee_testing/src/lib.rs
@@ -71,14 +71,14 @@ impl coffee_core::CoffeeArgs for CoffeeTestingArgs {
 /// we need to perform integration testing for coffee.
 pub struct CoffeeTesting {
     inner: CoffeeManager,
-    root_path: TempDir,
+    root_path: String,
 }
 
 impl Drop for CoffeeTesting {
     fn drop(&mut self) {
         use std::fs;
 
-        fs::remove_dir_all(self.root_path.path()).unwrap();
+        fs::remove_dir_all(self.root_path.to_owned()).unwrap();
     }
 }
 
@@ -96,7 +96,7 @@ impl CoffeeTesting {
             .map_err(|err| anyhow::anyhow!("{err}"))?;
         Ok(Self {
             inner: coffee,
-            root_path: dir,
+            root_path: dir.path().to_str().unwrap().to_owned(),
         })
     }
 
@@ -113,7 +113,7 @@ impl CoffeeTesting {
 
         Ok(Self {
             inner: coffee,
-            root_path: *tempdir,
+            root_path: tempdir.path().to_str().unwrap().to_owned(),
         })
     }

```

So I think this test needs a little bit more stress? maybe you should force to write to the store?

vincenzopalazzo:
This needs a rebase on master

vincenzopalazzo:
Ok Rebased on top of the current master

PR #162, opened 2023-07-04
feat(core): use the no sql database inside the coffee manager

vincenzopalazzo:
This adds more solid storage to coffee by using a no SQL database and using it to store the raw JSON.

However, there is a need to research if we need a more complex schema to make persistent the coffee information.

For now we keep it simple

### DIFF
--- a/coffee_storage/src/nosql_db.rs
+++ b/coffee_storage/src/nosql_db.rs
@@ -1,0 +1,51 @@
+use async_trait::async_trait;
+use coffee_lib::utils::check_dir_or_make_if_missing;
+use nosql_db::NoSQL;
+use nosql_sled::SledDB;
+
+use coffee_lib::error;
+use coffee_lib::errors::CoffeeError;
+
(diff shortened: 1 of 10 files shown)


coffee_storage/src/nosql_db.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// No SQL database
+
+pub struct NoSQlStorage {
+    inner: SledDB,
+}
+
+impl NoSQlStorage {
+    pub fn new(path: &str) -> Result<Self, CoffeeError> {
+        let db = SledDB::new(path).map_err(|err| error!("{err}"))?;
```
vincenzopalazzo, line 18:
```suggestion
       // make the storage directory inside the `path`
        let db = SledDB::new(format!("{}/storage").map_err(|err| error!("{err}"))?;
```

ok I just forget that the database needs the full directory

vincenzopalazzo: Done!

PR #164, opened 2023-07-05
feat(test): run the httpd inside the integration testing 

vincenzopalazzo:


coffee_testing/src/lib.rs
```diff
@@ -40,6 +39,36 @@ pub mod macros {
...
+            async {
+                use std::process::Stdio;
+
+                use tokio::process::Command;
+
+                let opt_args = format!($($opt_args)*);
+                let args = opt_args.trim();
+                let args_tok: Vec<&str> = args.split(" ").collect();
+
+                let cargo_target = std::env::var("CARGO_TARGET_DIR").unwrap();
```
vincenzopalazzo, line 54: this is wrong, I make this working in the https://github.com/coffee-tools/coffee/pull/91

vincenzopalazzo (commented):
a small self review

coffee_testing/src/lib.rs
vincenzopalazzo:
```rust
     let cargo_target = concat!(env!("CARGO_MANIFEST_DIR"), "/..");
    let path = std::path::Path::new(cargo_target).to_str().unwrap();
```

vincenzopalazzo (commented):
Ok this is the way to fix the httpd

vincenzopalazzo:
overrided by https://github.com/coffee-tools/coffee/pull/170

PR #169, opened 2023-07-08
Implement coffee HTTPD test framework 

tareknaser360:
This PR adds support for testing `coffee_httpd` crate

PR #170, opened 2023-07-09
add httpd testing framework

vincenzopalazzo:
Override https://github.com/coffee-tools/coffee/pull/169


vincenzopalazzo:
For now we consider the plugin test just a flacky tests

Issue #172, opened 2023-07-10
Unable to install the same plugin on 2 different networks 


tareknaser360:
Coffee should be able to work with different configurations based on different networks.

vincenzopalazzo:
We need also to share the repository with all the networks, currently we have a repository per network and it is wrong

PR #173, opened 2023-07-15
Deprecate the nurse command and add a test

tareknaser360:
e51181172ba221ed944917b1d0eb57aac82022fd modifies the current test framework to allow to run a lightning node in any network specified in the test itself. 
-  previously we could only run lightning in `regtest`

49548f0c247b19491aa56fcf41eba826bc14922f adds a new test to mimic the issue where coffee fails to install the same plugin in 2 different networks.

Related issue: #172

### DIFF
--- a/coffee_testing/src/cln.rs
+++ b/coffee_testing/src/cln.rs
@@ -72,26 +72,31 @@ impl Drop for Node {
 }
 
 impl Node {
-    pub async fn tmp() -> anyhow::Result<Self> {
-        Self::with_params("").await
+    pub async fn tmp(network: &str) -> anyhow::Result<Self> {
+        Self::with_params("", network).await
     }
 
-    pub async fn with_params(params: &str) -> anyhow::Result<Self> {
-        let btc = BtcNode::tmp().await?;
+    pub async fn with_params(params: &str, network: &str) -> anyhow::Result<Self> {
+        let btc = BtcNode::tmp(network).await?;
 
         let dir = tempfile::tempdir()?;
 
         let process = macros::lightningd!(
             dir,
             port::random_free_port().unwrap(),
-            "--network=regtest --log-level=debug --bitcoin-rpcuser={} --bitcoin-rpcpassword={} --bitcoin-rpcport={} {}",
+            "--network={} --log-level=debug --bitcoin-rpcuser={} --bitcoin-rpcpassword={} --bitcoin-rpcport={} {}",
+            network,
             btc.user,
             btc.pass,
             btc.port,
             params,
         )?;
 
-        let rpc = LightningRPC::new(dir.path().join(".lightning/regtest").join("lightning-rpc"));
+        let rpc = LightningRPC::new(
+            dir.path()
+                .join(format!(".lightning/{}", network))
+                .join("lightning-rpc"),
+        );
 
         wait_for!(async { rpc.getinfo() });
 
--- a/tests/src/coffee_httpd_integration_tests.rs
+++ b/tests/src/coffee_httpd_integration_tests.rs
@@ -11,7 +11,7 @@ use crate::init;
 pub async fn httpd_init_add_remote() {
     init();
 
-    let mut cln = Node::tmp().await.unwrap();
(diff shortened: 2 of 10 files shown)


coffee_testing/src/cln.rs
```diff
@@ -72,26 +72,31 @@ impl Drop for Node {
...
 impl Node {
-    pub async fn tmp() -> anyhow::Result<Self> {
-        Self::with_params("").await
+    pub async fn tmp(network: String) -> anyhow::Result<Self> {
+        Self::with_params("", network).await
     }
 
-    pub async fn with_params(params: &str) -> anyhow::Result<Self> {
-        let btc = BtcNode::tmp().await?;
+    pub async fn with_params(params: &str, network: String) -> anyhow::Result<Self> {
```
vincenzopalazzo, line 79: use `network: &str` so you can pass a literal string

tests/src/coffee_httpd_integration_tests.rs
```diff
@@ -11,7 +11,7 @@ use crate::init;
 pub async fn httpd_init_add_remote() {
     init();
 
-    let mut cln = Node::tmp().await.unwrap();
+    let mut cln = Node::tmp("regtest".to_string()).await.unwrap();
```
vincenzopalazzo, line 14: then you can use just `Node::tmp("regtest")`

tests/src/coffee_plugin_integration_tests.rs
```diff
@@ -12,7 +12,7 @@ pub async fn init_cln_with_coffee_plugin_test() {
     let path = std::path::Path::new(cargo_target).to_str().unwrap();
     let plugin_path = format!("{path}/target/debug/coffee_plugin");
     log::info!("plugin path {plugin_path}");
-    let mut cln = Node::with_params(&format!("--plugin={plugin_path}"))
+    let mut cln = Node::with_params(&format!("--plugin={plugin_path}"), "regtest".to_string())
```
vincenzopalazzo, line 15: same in this file

tests/src/coffee_integration_tests.rs
```diff
@@ -289,3 +289,78 @@ pub async fn test_errors_and_show() {
...
+    // initialize a lightning node in testnet network
+    let mut cln = Node::tmp("testnet".to_string()).await.unwrap();
+    let lightning_dir = cln.rpc().getinfo().unwrap().ligthning_dir;
+    let lightning_testnet_dir = lightning_dir.strip_suffix("/testnet").unwrap();
+    log::info!("lightning path for testnet network: {lightning_dir}");
+    let new_args = CoffeeTestingArgs {
+        conf: None,
+        data_dir: dir.path().clone().to_string_lossy().to_string(),
+        network: "testnet".to_string(),
+    };
```
vincenzopalazzo, line 345: we can assert on the dir? and make sure that you are running coffee on the same directory.

vincenzopalazzo (changes requested):
Very nice organisation and a lovely job.

I left some comments, also if there are any big issue

vincenzopalazzo:
This was the bug that we are looking for right?

```
coffee_1  | failures:
coffee_1  | 
coffee_1  | ---- coffee_integration_tests::install_plugin_in_two_networks stdout ----
coffee_1  | thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: CoffeeError { code: 1, msg: "plugin `summary` are not present inside the repositories" }', tests/src/coffee_integration_tests.rs:358:10
coffee_1  | stack backtrace:
coffee_1  |    0: rust_begin_unwind
```

tareknaser360:
> This was the bug that we are looking for right?

Yes. And by the way if we try to add `lightningd` as a remote repository for the `testnet`, it fails because it already exists in the `repositories` folder.
We need a way to separate coffee configurations based on network.

vincenzopalazzo:
>We need a way to separate coffee configurations based on network.

we need just a way to keep the coffee repositories network independent

vincenzopalazzo (approved):

PR #175, opened 2023-07-19
Implement the `show` endpoint in the httpd crate

tareknaser360:
- fixes the links to crates in the `README` file
- define a new type Show to send to the show http endpoint
- implement coffee_show function to handle /show endpoint
- add an httpd integration test to further stress test the server.

See commit body for details

### DIFF
--- a/coffee_httpd/src/httpd/server.rs
+++ b/coffee_httpd/src/httpd/server.rs
@@ -59,6 +59,7 @@ pub async fn run_httpd<T: ToSocketAddrs>(
             .service(coffee_remote_add)
             .service(coffee_remote_rm)
             .service(coffee_remote_list)
+            .service(coffee_show)
             .with_json_spec_at("/api/v1")
             .build()
     })
@@ -112,17 +113,7 @@ async fn coffee_remove(
 async fn coffee_list(data: web::Data<AppState>) -> Result<Json<Value>, Error> {
     let mut coffee = data.coffee.lock().await;
     let result = coffee.list().await;
-    match result {
-        Ok(coffee_list) => {
-            let val = serde_json::to_value(coffee_list).map_err(|err| {
-                actix_web::error::ErrorInternalServerError(format!("coffee list error: {err}"))
-            })?;
-            Ok(Json(val))
-        }
-        Err(err) => Err(actix_web::error::ErrorInternalServerError(format!(
-            "coffee list error: {err}"
-        ))),
-    }
+    handle_httpd_response!(result)
 }
 
 #[api_v2_operation]
@@ -163,19 +154,18 @@ async fn coffee_remote_list(data: web::Data<AppState>) -> Result<Json<Value>, Er
     let mut coffee = data.coffee.lock().await;
     let result = coffee.list_remotes().await;
 
-    match result {
-        Ok(coffee_remotes) => {
-            let val = serde_json::to_value(coffee_remotes).map_err(|err| {
-                actix_web::error::ErrorInternalServerError(format!(
-                    "Failed to list remote repositories: {err}"
-                ))
-            })?;
-            Ok(Json(val))
-        }
-        Err(err) => Err(actix_web::error::ErrorInternalServerError(format!(
-            "Failed to list remote repositories: {err}"
-        ))),
-    }
+    handle_httpd_response!(result)
+}
+
+#[api_v2_operation]
+#[get("/show")]
+async fn coffee_show(data: web::Data<AppState>, body: Json<Show>) -> Result<Json<Value>, Error> {
+    let plugin = &body.plugin;
+
+    let mut coffee = data.coffee.lock().await;
+    let result = coffee.show(plugin).await;
+
+    handle_httpd_response!(result)
 }
 
 // this is just a hack to support swagger UI with https://paperclip-rs.github.io/paperclip/
--- a/tests/src/coffee_httpd_integration_tests.rs
+++ b/tests/src/coffee_httpd_integration_tests.rs
@@ -67,3 +67,243 @@ pub async fn httpd_init_add_remote() {
 
     cln.stop().await.unwrap();
 }
+
+#[tokio::test(flavor = "multi_thread")]
+#[ntest::timeout(560000)]
+pub async fn httpd_add_remove_plugins() {
+    init();
+
+    let mut cln = Node::tmp("regtest").await.unwrap();
+    let lightning_dir = cln.rpc().getinfo().unwrap().ligthning_dir;
+    let lightning_dir = lightning_dir.strip_suffix("/regtest").unwrap();
+    let manager = CoffeeHTTPDTesting::tmp(lightning_dir.to_string()).await;
+    assert!(manager.is_ok(), "{:?}", manager);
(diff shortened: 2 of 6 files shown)


tests/src/coffee_httpd_integration_tests.rs
```diff
@@ -67,3 +67,241 @@ pub async fn httpd_init_add_remote() {
...
+#[ntest::timeout(560000)]
+pub async fn httpd_add_remove_plugins() {
+    init();
+
+    let mut cln = Node::tmp("regtest").await.unwrap();
+    let lightning_dir = cln.rpc().getinfo().unwrap().ligthning_dir;
+    let lightning_dir = lightning_dir.strip_suffix("/regtest").unwrap();
+    let manager = CoffeeHTTPDTesting::tmp(lightning_dir.to_string())
+        .await
+        .unwrap();
```
vincenzopalazzo, line 81: please do not unwrap, but assert on the result if it is ok and then unwrap later (if the assert is false, print the debug version of the error

The rule here is that you should unwrap only when you know that the error never happens in the test can happen, so please remove the unwrap and fix all occurrence in the test

tests/src/coffee_httpd_integration_tests.rs
```diff
@@ -67,3 +67,241 @@ pub async fn httpd_init_add_remote() {
...
+    log::info!("base url: {url}");
+    let client = reqwest::Client::new();
+
+    // Define the request body to be sent to the /remote/add endpoint
+    let remote_add_request = RemoteAdd {
+        repository_name: "lightningd".to_string(),
+        repository_url: "https://github.com/lightningd/plugins.git".to_string(),
+    };
+
+    // Send the request to add a remote repository
```
vincenzopalazzo, line 93: The comment is obvious so you can avoid to add the commit everywhere

The comment in the test is added when you are doing a not obvious things 

Change all occurrence in the tests

tests/src/coffee_httpd_integration_tests.rs
```diff
@@ -67,3 +67,241 @@ pub async fn httpd_init_add_remote() {
...
+    log::info!("/show response: {}", body);
+
+    // Parse the response body
+    let response_json: serde_json::Value = serde_json::from_str(&body).unwrap();
+
+    // Extract the `readme` field from the response JSON
+    let readme = response_json["readme"].as_str().unwrap();
+
+    // Assert that the `readme` starts with the expected content
+    assert!(readme.starts_with("# Helpme plugin"));
```
vincenzopalazzo, line 131:
```suggestion
    assert!(readme.starts_with("# Helpme plugin"), "{:?}", readme);
```

do always this inside an assert, if it is not a dubug add a Debug inside the struct derive

vincenzopalazzo (changes requested):
Ok I have added the review.

This is good but lets focus on the real refactoring that we need to do

We are not stressing in this way coffee at all

coffee_httpd/src/httpd/server.rs
```diff
@@ -178,6 +179,27 @@ async fn coffee_remote_list(data: web::Data<AppState>) -> Result<Json<Value>, Er
...
+        Ok(coffee_show) => {
+            let val = serde_json::to_value(coffee_show).map_err(|err| {
+                actix_web::error::ErrorInternalServerError(format!("Failure: {err}"))
+            })?;
+            Ok(Json(val))
+        }
+        Err(err) => Err(actix_web::error::ErrorInternalServerError(format!(
+            "Failure: {err}"
+        ))),
+    }
```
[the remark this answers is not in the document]
tareknaser360: The `handle_httpd_response` macro handles responses where there is a success message. 
It doesn't handle `list`, `remote/list` or `show` because these have a returned value that gets parsed in case the result is `ok`.

I defined another macro to handle responses for these endpoints

tests/src/coffee_httpd_integration_tests.rs
tareknaser360: Do you think this is a good use of `expect`?
We can
```rust
let manager = CoffeeHTTPDTesting::tmp(lightning_dir.to_string())
    .await
    .expect("Failed to start HTTPD");
```

coffee_httpd/src/httpd/server.rs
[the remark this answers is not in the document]
vincenzopalazzo: I do not think you need another macro. But you need just a macro that accepts the result content

tests/src/coffee_httpd_integration_tests.rs
vincenzopalazzo: No, this is the same. We are missing information about the error

```rust
let manager = CoffeeHTTPDTesting::tmp(lightning_dir.to_string())
    .await;
assert!(manager.is_ok(), "{:?}", manager);
// SAFETY: This is safe to unwrap because we check that it is Ok in the step before.
let manager = manager.unwrap();
```

Every codebase has code-style rules, which are mine for the test. Panic inside the test makes it hard to read the error when something wrong happens quickly. So in the case of the assert the test is able to point you in the correct direction.

In other codebases could you find that the unwrap is allowed, but I would like to disallow this everywhere also inside the test.

coffee_httpd/src/httpd/server.rs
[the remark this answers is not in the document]
tareknaser360: Ok I changed it to be 1 macro with 2 different conditions:
1- there is no return value in the result but we want to return a success message
2- there is a return value

Now we use the same macro in every endpoint in the server

tests/src/coffee_httpd_integration_tests.rs
```diff
@@ -67,3 +67,243 @@ pub async fn httpd_init_add_remote() {
...
+
+    // Log the response body
+    log::info!("/list response: {}", body);
+
+    // Assert that the "helpme" plugin was removed
+    let plugins = body["plugins"].as_array();
+    assert!(plugins.is_some(), "{:?}", plugins);
+    let plugins = plugins.unwrap();
+    assert!(
+        !plugins.iter().any(|plugin| plugin["name"] == "helpme"),
```
vincenzopalazzo, line 284:
```suggestion
        !(plugins.iter().any(|plugin| plugin["name"] == "helpme")),
```

tests/src/coffee_httpd_integration_tests.rs
```diff
@@ -67,3 +67,243 @@ pub async fn httpd_init_add_remote() {
...
+
+    // Log the response body
+    log::info!("/remote/list response: {}", body);
+
+    // Assert that the "lightningd" remote repository doesn't exist in the response
+    let remotes = body["remotes"].as_array();
+    assert!(remotes.is_some(), "{:?}", remotes);
+    let remotes = remotes.unwrap();
+    assert!(
+        !remotes
```
vincenzopalazzo, line 302: same here

vincenzopalazzo (changes requested):

vincenzopalazzo (commented):
LGTM, this needs only to fix the commit history in three

1. modify the macro
2. implement the endpoint
3. add testing for the endpoint

PR #176, opened 2023-07-20
supress intended clippy warnings 

tareknaser360:
- handles the `Err` variant in 2 statements in coffee.rs
- suppresses 2 clippy warnings that are intended 
- removes redundant `root_path` from `install_plugin_in_two_networks` test

### DIFF
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -90,18 +90,19 @@ impl CoffeeManager {
     /// when coffee is configured, run an inventory to collect all the necessary information
     /// about the coffee ecosystem.
     async fn inventory(&mut self) -> Result<(), CoffeeError> {
-        self.storage
+        let _ = self
+            .storage
             .load::<CoffeeStorageInfo>(&self.config.network)
             .await
-            .and_then(|store| {
+            .map(|store| {
                 self.config = store.config;
-                Ok(())
             });
         // FIXME: check if this exist in a better wai
-        self.storage
+        let _ = self
+            .storage
             .load::<HashMap<RepoName, RepositoryInfo>>("repositories")
             .await
-            .and_then(|item| {
+            .map(|item| {
                 log::debug!("repositories in store {:?}", item);
                 item.iter().for_each(|repo| match repo.1.kind {
                     Kind::Git => {
@@ -109,7 +110,6 @@ impl CoffeeManager {
                         self.repos.insert(repo.name(), Box::new(repo));
                     }
                 });
-                Ok(())
             });
 
         if let Err(err) = self.coffee_cln_config.parse() {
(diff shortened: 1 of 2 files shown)


coffee_core/src/coffee.rs
```diff
@@ -90,26 +90,26 @@ impl CoffeeManager {
     /// when coffee is configured, run an inventory to collect all the necessary information
     /// about the coffee ecosystem.
     async fn inventory(&mut self) -> Result<(), CoffeeError> {
-        self.storage
+        let _ = self
+            .storage
             .load::<CoffeeStorageInfo>(&self.config.network)
             .await
-            .and_then(|store| {
+            .map(|store| {
```
tareknaser360, line 97: This is a recommendation by clippy
```bash
warning: using `Result.and_then(|x| Ok(y))`, which is more succinctly expressed as `map(|x| y)`
```

coffee_core/src/coffee.rs
```diff
@@ -90,26 +90,26 @@ impl CoffeeManager {
...
                 self.config = store.config;
-                Ok(())
             });
```
vincenzopalazzo, line 99:
```suggestion
            })?;
```

Do you try to simply return the error?

vincenzopalazzo (changes requested):

coffee_core/src/coffee.rs
vincenzopalazzo: Clippy here is right. With `map` you are hiding possible errors, but in this case is not.

But I think the code can be easily solved by returning the error with `?`

coffee_core/src/coffee.rs
tareknaser360: Yes that is a cleaner way. thanks!

coffee_plugin/src/plugin/state.rs
```diff
@@ -41,6 +41,7 @@ impl State {
         self.args.clone().unwrap()
     }
 
+    #[allow(clippy::await_holding_lock)]
```
vincenzopalazzo, line 44: mh wait this is dangerus, what is clippy telling you?

vincenzopalazzo (changes requested):

coffee_plugin/src/plugin/state.rs
tareknaser360: This is the warning we discussed earlier over discord
the warning is 
```bash
warning: this `MutexGuard` is held across an `await` point
  --> coffee_plugin/src/plugin/state.rs:45:9
   |
45 | /         self.coffee()
46 | |             .lock()
47 | |             .unwrap()
   | |_____________________^
   |
   = help: consider using an async-aware `Mutex` type or ensuring the `MutexGuard` is dropped before calling await
note: these are all the `await` points this lock is held through
  --> coffee_plugin/src/plugin/state.rs:45:9
   |
45 | /         self.coffee()
46 | |             .lock()
47 | |             .unwrap()
48 | |             .setup(&self.args.clone().unwrap().conf)
49 | |             .await?;
   | |____________________^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#await_holding_lock
```
I couldn't solve it because it's used in `coffee_plugin/src/plugin/plugin_mod.rs` for example in regular synchronous functions so we can't use await  and we can't change them to `async` functions because they are functions in traits.

vincenzopalazzo (approved):
LGTM

PR #180, opened 2023-07-25
Install Plugins in a Separate Folder Specific to the Network

### LINKED ISSUE #178: Separate Directory for Plugin Installation to Enhance Version Control
Currently, coffee installs plugins in place, which may cause challenges in version management and updates. To optimize this process, we propose implementing a new approach:
- When cloning a remote repository locally, the latest commit of the repository will be stored for version tracking.
- Instead of installing the plugins directly in their original locations, a separate directory `.coffee/<network>/plugins` will be designated for plugin installation.

By adopting this solution, coffee will benefit from improved version control and easier plugin updates.

tareknaser360:
## Install Plugins in a Separate Directory:
coffee installs plugins in place, which could lead to challenges in version management and updates. To optimize this process, we have implemented a new approach:
1- When cloning a remote repository locally, we now store the latest commit of the repository to facilitate version tracking.
2- Instead of installing the plugins directly in their original locations, we have designated a separate directory `.coffee/<network>/plugins` for plugin installation.

This change improves version control and allows for easier plugin updates in the future.
Fixes #178

### DIFF
--- a/coffee_core/Cargo.toml
+++ b/coffee_core/Cargo.toml
@@ -6,6 +6,7 @@ edition = "2021"
 
 [dependencies]
 tokio = { version = "1.22.0", features = ["full"] }
+fs_extra = "1.3.0"
 async-trait = "0.1.57"
 coffee_lib = { path = "../coffee_lib" }
 coffee_github = { path = "../coffee_github"  }
--- a/coffee_core/Cargo.toml
+++ b/coffee_core/Cargo.toml
@@ -6,7 +6,6 @@ edition = "2021"
 
 [dependencies]
 tokio = { version = "1.22.0", features = ["full"] }
-fs_extra = "1.3.0"
 async-trait = "0.1.57"
 coffee_lib = { path = "../coffee_lib" }
 coffee_github = { path = "../coffee_github"  }
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -134,10 +134,10 @@ impl CoffeeManager {
         if let Some(rpc) = &self.rpc {
             let response = rpc
                 .send_request(method, payload)
-                .map_err(|err| CoffeeError::new(1, &format!("{err}")))?;
+                .map_err(|err| error!("{}", &format!("{err}")))?;
             log::debug!("cln answer with {:?}", response);
             if let Some(err) = response.error {
-                return Err(CoffeeError::new(1, &format!("cln error: {}", err.message)));
+                return Err(error!("{}", &format!("cln error: {}", err.message)));
             }
             return Ok(response.result.unwrap());
         }
@@ -204,7 +204,7 @@ impl CoffeeManager {
         self.load_cln_conf().await?;
         let mut conf = self.cln_config.clone().unwrap();
         conf.add_subconf(self.coffee_cln_config.clone())
-            .map_err(|err| CoffeeError::new(1, &err.cause))?;
+            .map_err(|err| error!("{}", &err.cause))?;
         conf.flush()?;
         Ok(())
     }
@@ -256,30 +256,21 @@ impl PluginManager for CoffeeManager {
                         plugin.exec_path = new_exec_path;
 
                         log::debug!("plugin: {:?}", plugin);
-                        let result = plugin.configure(verbose).await;
-                        log::debug!("result from plugin configure: {:?}", result);
-                        match result {
-                            Ok(path) => {
-                                log::debug!("runnable plugin path {path}");
-                                if !try_dynamic {
-                                    self.config.plugins.push(plugin);
-                                    log::debug!(
-                                        "path coffee conf: {}",
-                                        self.coffee_cln_config.path
-                                    );
-                                    self.coffee_cln_config
-                                        .add_conf("plugin", &path.to_owned())
-                                        .map_err(|err| CoffeeError::new(1, &err.cause))?;
-                                    log::debug!("coffee conf updated: {}", self.coffee_cln_config);
-                                    self.flush().await?;
-                                    self.update_conf().await?;
-                                } else {
-                                    self.start_plugin(&path).await?;
-                                }
-                                return Ok(());
-                            }
-                            Err(err) => return Err(err),
+                        let path = plugin.configure(verbose).await?;
+                        log::debug!("runnable plugin path {path}");
+                        if !try_dynamic {
+                            self.config.plugins.push(plugin);
+                            log::debug!("path coffee conf: {}", self.coffee_cln_config.path);
+                            self.coffee_cln_config
+                                .add_conf("plugin", &path.to_owned())
+                                .map_err(|err| error!("{}", err.cause))?;
+                            log::debug!("coffee conf updated: {}", self.coffee_cln_config);
+                            self.flush().await?;
+                            self.update_conf().await?;
+                        } else {
+                            self.start_plugin(&path).await?;
                         }
+                        return Ok(());
                     }
                     None => return Err(error!("exec path not found")),
                 };
@@ -302,7 +293,7 @@ impl PluginManager for CoffeeManager {
             log::debug!("coffee cln config: {}", self.coffee_cln_config);
             self.coffee_cln_config
                 .rm_conf("plugin", Some(&exec_path.to_owned()))
-                .map_err(|err| CoffeeError::new(1, &err.cause))?;
+                .map_err(|err| error!("{}", &err.cause))?;
             self.flush().await?;
             self.update_conf().await?;
             Ok(CoffeeRemove { plugin })
@@ -414,8 +405,8 @@ impl PluginManager for CoffeeManager {
                 return Ok(CoffeeShow { readme: contents });
             }
         }
-        let err = CoffeeError::new(
-            1,
+        let err = error!(
+            "{}",
             &format!("plugin `{plugin}` are not present inside the repositories"),
         );
         Err(err)
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -1,6 +1,5 @@
 //! Coffee mod implementation
 use coffee_storage::nosql_db::NoSQlStorage;
-use fs_extra::dir;
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::vec::Vec;
@@ -13,14 +12,15 @@ use clightningrpc_conf::{CLNConf, SyncCLNConf};
 use log;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
+use tokio::process::Command;
 
 use coffee_github::repository::Github;
-use coffee_lib::error;
 use coffee_lib::errors::CoffeeError;
 use coffee_lib::plugin_manager::PluginManager;
 use coffee_lib::repository::Repository;
 use coffee_lib::types::response::*;
 use coffee_lib::url::URL;
+use coffee_lib::{error, sh};
 use coffee_storage::model::repository::{Kind, Repository as RepositoryInfo};
 use coffee_storage::storage::StorageManager;
 
@@ -188,7 +188,7 @@ impl CoffeeManager {
         log::info!("looking for the cln config: {path}");
         file.parse()
             .map_err(|err| CoffeeError::new(err.core, &err.cause))?;
-        log::debug!("{:#?}", file.fields);
+        log::trace!("{:?}", file.fields);
         self.cln_config = Some(file);
         Ok(())
     }
@@ -227,7 +227,7 @@ impl PluginManager for CoffeeManager {
         // keep track if the plugin is successfully installed
         for repo in self.repos.values() {
             if let Some(mut plugin) = repo.get_plugin_by_name(plugin) {
-                log::trace!("{:#?}", plugin);
+                log::trace!("{:?}", plugin);
 
                 // old_root_path is the path where the plugin is cloned and currently stored
                 // eg. ~/.coffee/repositories/<repo_name>/<plugin_name>
@@ -240,13 +240,19 @@ impl PluginManager for CoffeeManager {
                     self.config.network,
                     plugin.name()
                 );
-                dir::copy(
-                    &old_root_path,
-                    format!("{}/{}/plugins/", self.config.root_path, self.config.network),
-                    &dir::CopyOptions::new(),
-                )
-                .map_err(|err| error!("{}", err.to_string()))?;
 
+                log::debug!(
+                    "Start! copying directory from {} inside the new one {}",
+                    old_root_path,
+                    new_root_path
+                );
+                let script = format!("cp -r {old_root_path} {new_root_path}");
+                sh!(self.config.root_path.clone(), script, verbose);
+                log::debug!(
+                    "Done! copying directory from {} inside the new one {}",
+                    old_root_path,
+                    new_root_path
+                );
                 let old_exec_path = plugin.exec_path.clone();
 
                 match old_exec_path.strip_prefix(&old_root_path) {
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -287,7 +287,12 @@ impl PluginManager for CoffeeManager {
         if let Some(index) = plugins.iter().position(|x| x.name() == plugin) {
             let plugin = plugins[index].clone();
             let exec_path = plugin.exec_path.clone();
-            fs::remove_dir_all(plugin.root_path.clone()).await?;
+            let root_path = plugin.root_path.clone();
+            let cloned_repositories_path = format!("{}/repositories", self.config.root_path,);
+            // make sure that we are not deleting the cloned repositories
+            if !root_path.contains(&cloned_repositories_path) {
+                fs::remove_dir_all(root_path).await?;
+            }
             log::debug!("runnable plugin path: {exec_path}");
             plugins.remove(index);
             log::debug!("coffee cln config: {}", self.coffee_cln_config);
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -1,5 +1,6 @@
 //! Coffee mod implementation
 use coffee_storage::nosql_db::NoSQlStorage;
+use fs_extra::dir;
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::vec::Vec;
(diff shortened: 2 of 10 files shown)


coffee_core/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 
 [dependencies]
 tokio = { version = "1.22.0", features = ["full"] }
+fs_extra = "1.3.0"
```
vincenzopalazzo, line 9: what is this?

coffee_core/src/coffee.rs
```diff
@@ -223,7 +224,23 @@ impl PluginManager for CoffeeManager {
...
+
+                let source_folder = plugin.root_path.clone();
+                let destination_folder =
+                    format!("{}/{}/plugins", self.config.root_path, self.config.network);
+                dir::copy(
+                    source_folder,
+                    destination_folder.clone(),
+                    &dir::CopyOptions::new(),
+                )
+                .map_err(|err| error!("{}", err.to_string()))?;
```
vincenzopalazzo, line 236: This is not `async` so we are losing the async by blocking this call. It is tricky, but we must write out a version.

Take as an example this code and implement it with tokio

```rust
/// Copy files from source to destination recursively.
pub fn copy_recursively(source: impl AsRef<Path>, destination: impl AsRef<Path>) -> io::Result<()> {
    fs::create_dir_all(&destination)?;
    for entry in fs::read_dir(source)? {
        let entry = entry?;
        let filetype = entry.file_type()?;
        if filetype.is_dir() {
            copy_recursively(entry.path(), destination.as_ref().join(entry.file_name()))?;
        } else {
            fs::copy(entry.path(), destination.as_ref().join(entry.file_name()))?;
        }
    }
    Ok(())
}
```

We have already some utils in the coffee libs, we may wan put another function in there too

coffee_core/src/coffee.rs
```diff
@@ -223,7 +224,23 @@ impl PluginManager for CoffeeManager {
...
+
+                let result = plugin
+                    .configure(
+                        &format!("{}/{}", &destination_folder, plugin.name()).to_string(),
+                        verbose,
+                    )
+                    .await;
```
vincenzopalazzo, line 243:
```suggestion
                    .await?;
```

and then manage only the path, in this way we simplify the code

coffee_core/src/coffee.rs
```diff
@@ -256,6 +273,14 @@ impl PluginManager for CoffeeManager {
         let plugins = &mut self.config.plugins;
         if let Some(index) = plugins.iter().position(|x| x.name() == plugin) {
             let plugin = plugins[index].clone();
+            let root_path = format!(
+                "{}/{}/plugins/{}",
+                self.config.root_path,
+                self.config.network,
+                plugin.name()
+            );
```
vincenzopalazzo, line 281: this will be `plugin.root_path`, you do not need to rebuild the path each time, the plugin should have this information already

coffee_github/src/repository.rs
```diff
@@ -36,6 +36,8 @@ pub struct Github {
     /// the name of the branch to be able to
     /// pull the changes from the correct branch
     branch: String,
+    /// the commit id of the last commit
+    commit: String,
```
vincenzopalazzo, line 40:
```suggestion
    commit: Option<String>,
```

When we are restoring the object from the database we should not crash (if someone is using coffee before this change) so we use optional and then after a while we change it is required (put a commit on it as a FIXME).

coffee_github/src/repository.rs
```diff
@@ -57,6 +59,7 @@ impl Github {
             url: url.clone(),
             plugins: vec![],
             branch: "".to_owned(),
+            commit: "".to_owned(),
```
vincenzopalazzo, line 62: Mh btw this is wrong for this PR I think? we should store the commit inside the plugin, we already know the commit of the repository we are using git for this reason

coffee_github/src/repository.rs
```diff
@@ -210,6 +213,13 @@ impl Repository for Github {
                 } else {
                     "main".to_owned()
                 };
+                self.commit = repo
+                    .head()
+                    .map_err(|err| error!("{}", err.message()))?
+                    .peel_to_commit()
+                    .map_err(|err| error!("{}", err.message()))?
+                    .id()
+                    .to_string();
```
vincenzopalazzo, line 222: please write a macros for this named `git_head!` :)

coffee_github/src/repository.rs
```diff
@@ -318,6 +331,7 @@ impl From<&Github> for StorageRepository {
             url: value.url.to_owned(),
             plugins: value.plugins.to_owned(),
             branch: value.branch.to_owned(),
+            commit: value.commit.to_owned(),
```
vincenzopalazzo, line 334: yeah nice but it is the wrong place, you can always access to git to extract this information, the commit should be inside the plugin when we move out the plugin in another directory

coffee_lib/src/plugin.rs
```diff
@@ -125,19 +126,24 @@ impl Plugin {
     /// configure the plugin in order to work with cln.
     ///
     /// In case of success return the path of the executable.
-    pub async fn configure(&mut self, verbose: bool) -> Result<String, CoffeeError> {
+    pub async fn configure(
+        &mut self,
+        root_path: &str,
```
vincenzopalazzo, line 131:
```suggestion
```

Should be already inside the plugin, please take a look on how coffee index the plugins

coffee_lib/src/plugin.rs
```diff
@@ -125,19 +126,24 @@ impl Plugin {
...
+        verbose: bool,
+    ) -> Result<String, CoffeeError> {
         let exec_path = if let Some(conf) = &self.conf {
             if let Some(script) = &conf.plugin.install {
-                sh!(self.root_path.clone(), script, verbose);
+                eprintln!("installing plugin...");
```
vincenzopalazzo, line 136:
```suggestion
```

we have the logger for this, but we do not need it

coffee_lib/src/plugin.rs
```diff
@@ -125,19 +126,24 @@ impl Plugin {
...
+                sh!(root_path, script, verbose);
                 self.exec_path.clone()
             } else {
                 self.lang
-                    .default_install(&self.root_path, &self.name, verbose)
```
vincenzopalazzo, line 135: this was correct

coffee_lib/src/plugin.rs
```diff
@@ -125,19 +126,24 @@ impl Plugin {
...
+                    .default_install(root_path, &self.name, verbose)
                     .await?
             }
         } else {
             self.lang
-                .default_install(&self.root_path, &self.name, verbose)
```
vincenzopalazzo, line 140: same here

coffee_storage/src/model/repository.rs
```diff
@@ -15,4 +15,5 @@ pub struct Repository {
     pub url: URL,
     pub plugins: Vec<Plugin>,
     pub branch: String,
+    pub commit: String,
```
vincenzopalazzo, line 18: wrong place

vincenzopalazzo (changes requested):
Ok some comments to help you understand my idea

coffee_lib/src/plugin.rs
tareknaser360: yes that slipped
I didn't mean to push this

coffee_lib/src/macros.rs
```diff
@@ -48,4 +48,18 @@ macro_rules! sh {
...
+            .head()
+            .map_err(|err| format!("{}", err.message()))?
+            .peel_to_commit()
+            .map_err(|err| format!("{}", err.message()))?
+            .id()
+            .to_string()
+    }};
+}
+
+pub use {commit_id, error, sh};
```
vincenzopalazzo, line 65:
```suggestion
pub use commit_id, error, sh;
```

coffee_lib/src/plugin.rs
```diff
@@ -112,13 +117,17 @@ impl Plugin {
         path: &str,
         plugin_lang: PluginLang,
         config: Option<Conf>,
+        commit_id: Option<String>,
+        network: Option<String>,
```
vincenzopalazzo, line 121: we do not need the network, or better we MUST not need

vincenzopalazzo (changes requested):

coffee_lib/src/plugin.rs
tareknaser360: this plugin struct gets first initialized when the remote repo is cloned locally and the plugins are not yet installed in any network
so until it's installed it's going to be `None`

coffee_lib/src/macros.rs
tareknaser360: This doesn't work for me locally at least

coffee_core/src/coffee.rs
```diff
@@ -223,27 +224,61 @@ impl PluginManager for CoffeeManager {
...
+                                    log::debug!("coffee conf updated: {}", self.coffee_cln_config);
+                                    self.flush().await?;
+                                    self.update_conf().await?;
+                                } else {
+                                    self.start_plugin(&path).await?;
+                                }
+                                return Ok(());
+                            }
+                            Err(err) => return Err(err),
                         }
```
vincenzopalazzo, line 278: lets clean up a little bit this code maybe in a new commit
```suggestion
                        let path = plugin.configure(verbose).await?;
                        log::debug!("runnable plugin path {path}");
                         if !try_dynamic {
                                    self.config.plugins.push(plugin);
                                    log::debug!(
                                        "path coffee conf: {}",
                                        self.coffee_cln_config.path
                                    );
                                    self.coffee_cln_config
                                        .add_conf("plugin", &path.to_owned())
                                        .map_err(|err| CoffeeError::new(1, &err.cause))?;
                                    log::debug!("coffee conf updated: {}", self.coffee_cln_config);
                                    self.flush().await?;
                                    self.update_conf().await?;
                                } else {
                                    self.start_plugin(&path).await?;
                                }
                                return Ok(());
                            }
```

ok on github this is hard, but you can keep just refactoring it :)

coffee_core/src/coffee.rs
```diff
@@ -278,6 +314,8 @@ impl PluginManager for CoffeeManager {
     }
 
     async fn upgrade(&mut self, repo: &str) -> Result<CoffeeUpgrade, CoffeeError> {
+        // TODO: upgrade should now be able to upgrade a single plugin
+        // without affecting other plugins installed from the same repo
```
vincenzopalazzo, line 318: yeah now there is a very good moment

coffee_github/src/repository.rs
```diff
@@ -10,7 +10,7 @@ use walkdir::DirEntry;
 use walkdir::WalkDir;
 
 use coffee_lib::errors::CoffeeError;
-use coffee_lib::macros::error;
+use coffee_lib::macros::{commit_id, error};
```
vincenzopalazzo, line 13: in rust there is a not written concept of prelude, where you can do `coffee_lib::prelude::*` and this include all the default configuration of it

I do not know if this will be the case, but I think if this macros start to grow we can think to make something like this

coffee_github/src/repository.rs
```diff
@@ -68,6 +68,9 @@ impl Github {
             .max_depth(1)
             .into_iter()
             .filter_entry(|dir_entry| !is_hidden(dir_entry));
+        let repo = git2::Repository::open(&self.url.path_string)
+            .map_err(|err| CoffeeError::new(1, err.message()))?;
```
vincenzopalazzo, line 72:
```suggestion
            .map_err(|err| error!(err.message()))?;
```

coffee_lib/src/macros.rs
```diff
@@ -48,4 +48,18 @@ macro_rules! sh {
...
-pub use {error, sh};
+/// get the repository's commit ID as a string.
+#[macro_export]
+macro_rules! commit_id {
+    ($repo:expr) => {{
+        $repo
+            .head()
+            .map_err(|err| format!("{}", err.message()))?
+            .peel_to_commit()
+            .map_err(|err| format!("{}", err.message()))?
```
vincenzopalazzo, line 59:
```suggestion
            .map_err(|err| error!("{}", err.message()))?
            .peel_to_commit()
            .map_err(|err| error!("{}", err.message()))?
```

coffee_lib/src/macros.rs
vincenzopalazzo: hahah crazy, but ok haha

vincenzopalazzo (commented):
The first look is very nice and clean. I left some comments, then I need to test it

I would also write a test to ensure that the logic is sane, and the test will go to install the plugin and check the coffee.conf contains the right path, and the plugin exists in this new path.

Then I will remove the plugin and see if the plugin is deleted inside this new path but still exist inside the old path.

P.S: I do not know if, with all the information that we expose, you can write the test :)

coffee_github/src/repository.rs
tareknaser360: I think for now at least we don't need this?
if we do `use coffee_lib::macros::*;` we would have to change every occurrence of `error!` to `macros::error!`

vincenzopalazzo: yeah just something to keep in mind

coffee_core/src/coffee.rs
```diff
@@ -257,12 +283,13 @@ impl PluginManager for CoffeeManager {
         if let Some(index) = plugins.iter().position(|x| x.name() == plugin) {
             let plugin = plugins[index].clone();
             let exec_path = plugin.exec_path.clone();
+            fs::remove_dir_all(plugin.root_path.clone()).await?;
```
vincenzopalazzo, line 286: You should check if the root path is not the repository path, in this way you are corrupting the repository for the current user of coffee (just mee?)

So here you need to check if the repository path of the plugin is == to the plugin.root path (in some way)

vincenzopalazzo (changes requested):

coffee_core/src/coffee.rs
tareknaser360: Do you mean to check if `plugin.root_path` is not in `.coffee/repositories` so not to remove the cloned version?

vincenzopalazzo: if the root path is equal to the `{repository_root_path}/{plugin_name}` this means that you are inside the repository and the plugin is not installed with the new procedure. So you are deleting the repository and corrupting it

vincenzopalazzo: >Do you mean to check if plugin.root_path is not in .coffee/repositories so not to remove the cloned version?

This is true for new plugins but what happens if you have coffee that already indexing and install a plugin with the current master version?

If I ran you the `coffee remove` with my current setup coffee delete the directory inside the repository because you are not thinking to be retro compatible with older version of coffee

coffee_core/src/coffee.rs
```diff
@@ -227,17 +228,42 @@ impl PluginManager for CoffeeManager {
...
+                    self.config.root_path,
+                    self.config.network,
+                    plugin.name()
+                );
+                dir::copy(
+                    &old_root_path,
+                    format!("{}/{}/plugins/", self.config.root_path, self.config.network),
+                    &dir::CopyOptions::new(),
+                )
+                .map_err(|err| error!("{}", err.to_string()))?;
```
vincenzopalazzo, line 248:
```suggestion
                .map_err(|err| error!("{err}")))?;
```

coffee_core/src/coffee.rs
```diff
@@ -227,17 +228,42 @@ impl PluginManager for CoffeeManager {
...
+
+                let old_exec_path = plugin.exec_path.clone();
+
```
vincenzopalazzo, line 251:
```suggestion
```

coffee_core/src/coffee.rs
```diff
@@ -261,12 +287,18 @@ impl PluginManager for CoffeeManager {
         if let Some(index) = plugins.iter().position(|x| x.name() == plugin) {
             let plugin = plugins[index].clone();
             let exec_path = plugin.exec_path.clone();
+            let root_path = plugin.root_path.clone();
+            let cloned_repositories_path = format!("{}/repositories", self.config.root_path,);
```
vincenzopalazzo, line 291:
```suggestion
            let cloned_repositories_path = format!("{}/repositories", self.config.root_path);
```

vincenzopalazzo (changes requested):
Some minor changes, but I found a bug in the remove.

So let's imagine that I had folgore installed on my system with the master version of Coffee. Now I want to run the remove of coffee and install it again, but the remove is not working.

This is the reproducible

```
➜  coffee git:(pr180-2) ✗ coffee --network testnet list
● Plugin installed
╭───────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ●   Language   Name      Exec path                                                                    │
├───────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ ●   rust       folgore   /home/vincent/.coffee/repositories/folgore-git/target/release/folgore_plugin │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────╯
➜  coffee git:(pr180-2) ✗ coffee --network testnet remove folgore
✗ Uninstalling plugin folgore error: Error while uninstalling the plugin: code: 1, msg: field with `plugin` not present

```

I guess is that we are trying to remove from the coffee configuration the plugin with the wrong path :/

I propose a suggestion too

coffee_core/src/coffee.rs
```diff
@@ -261,12 +287,18 @@ impl PluginManager for CoffeeManager {
...
+            // make sure that we are not deleting the cloned repositories
+            if !root_path.contains(&cloned_repositories_path) {
+                fs::remove_dir_all(root_path).await?;
+            }
```
vincenzopalazzo, line 295: This is not enough, if we are in the repository this means that we are running an old version of coffee and we should also remove the plugin from the configuration with the old path.

We should change the following code to work in this if else and remove the correct path

```
            self.coffee_cln_config
                .rm_conf("plugin", Some(&exec_path.to_owned()))
                .map_err(|err| CoffeeError::new(1, &err.cause))?;
                .map_err(|err| error!("{}", &err.cause))?;
```

vincenzopalazzo (commented):
I foget to push the suggestion

coffee_core/src/coffee.rs
tareknaser360: I can't fully understand the problem here
There are 2 cases:
- we are running the newest version of coffee --> no problem here
- we are running an older version of coffee:
      -  we only need to avoid removing the `root_path` of the plugin because that would be the cloned version.
      -  there is no problem with `exec_path` really because this path will be the same to the one we added to `cln_config` before. coffee doesn't change `exec_path` at any point in time. the new update only modifies it before adding it to `cln_config` normally. 
      
Am I missing something here?

tareknaser360: Also, it would be hard to craft the `exec_path` of a plugin instead of just using it from the field directly.
it would look something like `.coffee/repositories/<repo_name>/<plugin_name>` but we don't store the name of the repository from which the plugin was installed. (I think we should in the future)

vincenzopalazzo: Ah my bad hahahhahha

```
[2023-07-31T05:27:35Z DEBUG coffee_core::coffee] coffee cln config: # core lightning configuration generated by coffe please do not edit this
    #plugin=/home/vincent/.coffee/repositories/folgore-git/target/release/folgore_plugin
    
    
✗ Uninst
```

look at the `#`, sorry I was sleeping yesterday,

vincenzopalazzo (commented):
```
[2023-07-31T05:39:34Z DEBUG coffee_core::coffee] plugin: Plugin { name: "folgore", root_path: "/home/vincent/.coffee/testnet/plugins/folgore", exec_path: "/home/vincent/.coffee/testnet/plugins/folgore/target/release/folgore_plugin", lang: Rust, conf: Some(Conf { plugin: Plugin { name: "folgore", version: "0.0.1", lang: "rust", deprecated: None, dependencies: None, install: Some("make coffee\n"), main: "target/release/folgore_plugin" } }), commit: None }
[2023-07-31T05:39:34Z DEBUG coffee_lib::plugin] cmds: Split(SplitInternal { start: 0, end: 11, matcher: StrSearcher { haystack: "make coffee", needle: "\n", searcher: TwoWay(TwoWaySearcher { crit_pos: 0, crit_pos_back: 1, period: 1, byteset: 1024, position: 0, end: 11, memory: 0, memory_back: 1 }) }, allow_trailing_empty: true, finished: false })
[2023-07-31T05:39:34Z DEBUG coffee_lib::plugin] cmd "make coffee"
thread 'main' panicked at 'Unable to run the command: Os { code: 2, kind: NotFound, message: "No such file or directory" }', /home/vincent/Github/coffee/coffee/coffee_lib/src/plugin.rs:135:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[2023-07-31T05:39:34Z DEBUG sled::pagecache::logger] IoBufs dropped
```

This is the error that I am getting while instilling folgore with coffee.

Then we do not clean up the directory in case of failure, so  if I rerun `coffee install folgore` we try as monkey to copy the directory again.

```
[2023-07-31T05:44:13Z DEBUG coffee_core::coffee] finish plugin manager inventory
[2023-07-31T05:44:13Z DEBUG coffee_core::coffee] installing plugin: folgore
✗ code: 1, msg: Path "/home/vincent/.coffee/testnet/plugins/folgore-git/client.rs" exists
[2023-07-31T05:44:13Z DEBUG sled::pagecache::logger] IoBufs dropped

```

Please do some real tests on your environment and use coffee. Otherwise, we will never end this command. and we MUST not break user otherwise, coffee is useless

coffee_core/src/coffee.rs
vincenzopalazzo: Please see the fix https://github.com/coffee-tools/coffee/pull/180/commits/7edbc1bbb462a4a9f78e61726616fb192eedce19 this also make the copy async and more easy to read

This should work also for the double copy of the same dir two time :) Easy

tareknaser360: Nice. Thanks!

vincenzopalazzo (approved):

PR #191, opened 2023-08-02
Modify `CoffeeRemoteList` to store the git HEAD 

### LINKED ISSUE #186: Store the git HEAD information on the repository
We should store and display the git HEAD for each repo and print it in a table column in the terminal.

Maybe also the date of the last commit would be useful to display too

tareknaser:
Fixes #186

### DIFF
--- a/coffee_cmd/src/coffee_term/command_show.rs
+++ b/coffee_cmd/src/coffee_term/command_show.rs
@@ -54,15 +54,27 @@ pub fn show_remote_list(remote_list: Result<CoffeeRemote, CoffeeError>) -> Resul
         term::format::bold(String::from("Repository Alias")),
         term::format::bold(String::from("URL")),
         term::format::bold(String::from("N. Plugins")),
+        term::format::bold(String::from("Git HEAD")),
+        term::format::bold(String::from("Last Update")),
     ]);
     table.divider();
 
     for repository in &repositories {
+        let commit_id = match &repository.commit_id {
+            Some(commit_id) => commit_id.to_owned(),
+            None => String::from(""),
+        };
+        let date = match &repository.date {
+            Some(date) => date.to_owned(),
+            None => String::from(""),
+        };
         table.push([
             term::format::positive("●").into(),
             term::format::highlight(repository.local_name.to_owned()),
             term::format::bold(repository.url.to_owned()),
             term::format::highlight(repository.plugins.len().to_string()),
+            term::format::primary(commit_id),
+            term::format::bold(date),
         ])
     }
     table.print();
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -73,7 +73,7 @@ pub struct CoffeeManager {
     /// core lightning rpc connection
     rpc: Option<Client>,
     /// Recovery Strategies for the nurse command.
-    recovety_strategies: RecoveryChainOfResponsibility,
+    recovery_strategies: RecoveryChainOfResponsibility,
 }
 
 impl CoffeeManager {
@@ -86,7 +86,7 @@ impl CoffeeManager {
             storage: NoSQlStorage::new(&conf.root_path).await?,
             cln_config: None,
             rpc: None,
-            recovety_strategies: RecoveryChainOfResponsibility::new().await?,
+            recovery_strategies: RecoveryChainOfResponsibility::new().await?,
         };
         coffee.inventory().await?;
         Ok(coffee)
@@ -437,7 +437,7 @@ impl PluginManager for CoffeeManager {
     }
 
     async fn nurse(&mut self) -> Result<(), CoffeeError> {
-        self.recovety_strategies.scan().await
+        self.recovery_strategies.scan().await
     }
 }
 
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -14,13 +14,14 @@ use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use tokio::process::Command;
 
+use chrono::{TimeZone, Utc};
 use coffee_github::repository::Github;
 use coffee_lib::errors::CoffeeError;
 use coffee_lib::plugin_manager::PluginManager;
 use coffee_lib::repository::Repository;
 use coffee_lib::types::response::*;
 use coffee_lib::url::URL;
-use coffee_lib::{error, sh};
+use coffee_lib::{commit_id, error, get_repo_info, sh};
 use coffee_storage::model::repository::{Kind, Repository as RepositoryInfo};
 use coffee_storage::storage::StorageManager;
 
@@ -324,7 +325,7 @@ impl PluginManager for CoffeeManager {
         // without affecting other plugins installed from the same repo
         let repository = self
             .repos
-            .get(repo)
+            .get_mut(repo)
             .ok_or_else(|| error!("Repository with name: `{}` not found", repo))?;
 
         let status = repository.upgrade(&self.config.plugins).await?;
@@ -396,10 +397,15 @@ impl PluginManager for CoffeeManager {
     async fn list_remotes(&mut self) -> Result<CoffeeRemote, CoffeeError> {
         let mut remote_list = Vec::new();
         for repo in self.repos.values() {
+            let repository = git2::Repository::open(repo.url().path_string.as_str())
+                .map_err(|err| error!("{}", err.message()))?;
+            let (commit, date) = get_repo_info!(repository);
             remote_list.push(CoffeeListRemote {
                 local_name: repo.name(),
                 url: repo.url().url_string,
                 plugins: repo.list().await?,
+                commit_id: Some(commit),
+                date: Some(date),
             });
         }
         Ok(CoffeeRemote {
--- a/coffee_github/Cargo.toml
+++ b/coffee_github/Cargo.toml
@@ -13,3 +13,4 @@ log = "0.4.17"
 env_logger = "0.9.3"
 serde_yaml = "^0.9.0"
 walkdir = "2.3.2"
+chrono = { version = "0.4", features = ["std"], default-features = false }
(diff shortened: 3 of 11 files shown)


vincenzopalazzo:
The solution is inside the error

```
coffee_1  |    Compiling tests v0.1.0 (/workdir/tests)
coffee_1  | error[E0283]: type annotations needed
coffee_1  |   --> tests/src/logger.rs:42:21
coffee_1  |    |
coffee_1  | 42 |                     DateTime::from(SystemTime::now())
coffee_1  |    |                     ^^^^^^^^^^^^^^ cannot infer type for struct `chrono::DateTime<_>`
coffee_1  |    |
coffee_1  |    = note: multiple `impl`s satisfying `chrono::DateTime<_>: From<SystemTime>` found in the `chrono` crate:
coffee_1  |            - impl From<SystemTime> for chrono::DateTime<Local>;
coffee_1  |            - impl From<SystemTime> for chrono::DateTime<Utc>;
coffee_1  | 
coffee_1  | For more information about this error, try `rustc --explain E0283`.
coffee_1  | error: could not compile `tests` (lib test) due to previous error
coffee_1  | make: *** [Makefile:35: integration] Error 101
```

But this depends from your code, so I think you should just play a little bit of what type `SystemTime::now()` return

coffee_cmd/src/coffee_term/command_show.rs
```diff
@@ -54,6 +54,8 @@ pub fn show_remote_list(remote_list: Result<CoffeeRemote, CoffeeError>) -> Resul
         term::format::bold(String::from("Repository Alias")),
         term::format::bold(String::from("URL")),
         term::format::bold(String::from("N. Plugins")),
+        term::format::bold(String::from("Git HEAD")),
+        term::format::bold(String::from("Date of last commit")),
```
vincenzopalazzo, line 58: s/"Date of last commit"/"last update"

It is better `Date of last commit`

coffee_core/src/coffee.rs
```diff
@@ -396,10 +396,20 @@ impl PluginManager for CoffeeManager {
...
         let mut remote_list = Vec::new();
         for repo in self.repos.values() {
+            let commit_id = match repo.git_head() {
+                Some(val) => val,
+                None => String::from(""),
+            };
+            let date = match repo.date_of_the_last_commit() {
+                Some(val) => val,
+                None => String::from(""),
+            };
```
vincenzopalazzo, line 406: Why do not store the date as a string if we have to store something, or if it is None we keep it None?

In rust there is the optional type so we should use it

coffee_github/Cargo.toml
```diff
@@ -13,3 +13,4 @@ log = "0.4.17"
 env_logger = "0.9.3"
 serde_yaml = "^0.9.0"
 walkdir = "2.3.2"
+chrono = { version = "0.4", features = ["std"], default-features = false }
```
vincenzopalazzo, line 16: try to use this https://crates.io/crates/localtime/1.3.0

coffee_lib/src/repository.rs
```diff
@@ -33,5 +33,14 @@ pub trait Repository: Any {
...
 
+    /// return the git head of the repository.
+    fn git_head(&self) -> Option<String>;
+
+    /// return the date of the last commit of the repository.
+    fn date_of_the_last_commit(&self) -> Option<String>;
+
+    /// return the git information of the repository.
+    async fn git_info(&self) -> Result<(String, String), CoffeeError>;
+
```
vincenzopalazzo: If you noted inside the repository we do not have any reference to git, so the coffee_lib specifies only the central concept of the repository, but the repository is not coupled with git at all.

The git implementation is the coffee_github one, and where these methods should live too

coffee_github/src/repository.rs
```diff
@@ -221,9 +219,7 @@ impl Repository for Github {
                 } else {
                     "main".to_owned()
                 };
-                let (git_head, date_of_the_last_commit) = self.git_info().await?;
-                self.git_head = Some(git_head);
-                self.date_of_the_last_commit = Some(date_of_the_last_commit);
+                self.git_info().await?;
```
vincenzopalazzo, line 222: You do not need to have the method of the object, but the procedure can be static (without self) or just a macros, and this get_info method can return the information about the commit header and the commit time

coffee_github/src/repository.rs
```diff
@@ -255,6 +251,9 @@ impl Repository for Github {
         }
         // pull the changes from the repository
         let status = fast_forward(&self.url.path_string, &self.branch)?;
+        // update the git information
+        // (git HEAD and date of the last commit)
+        self.git_info().await?;
```
vincenzopalazzo, line 256: I would like to see something like `let (commit, time) = get_repo_info!()`  and then so just call, `self.commit = commit`

vincenzopalazzo (changes requested):
With the concept we are there but I think the implementation si mixing different concepts

coffee_core/src/coffee.rs
vincenzopalazzo: remove it and store inside the `CoffeeListRemote` the optional value

coffee_github/src/repository.rs
```diff
@@ -36,6 +37,10 @@ pub struct Github {
     /// the name of the branch to be able to
     /// pull the changes from the correct branch
     branch: String,
+    /// the latest commit id of the repository
+    git_head: Option<String>,
+    /// the latest commit date of the repository
+    date_of_the_last_commit: Option<String>,
```
vincenzopalazzo, line 43: call it just `last_activity`

coffee_lib/src/repository.rs
```diff
@@ -33,5 +33,14 @@ pub trait Repository: Any {
     /// return the url of the repository.
     fn url(&self) -> URL;
 
+    /// return the git head of the repository.
+    fn git_head(&self) -> Option<String>;
+
+    /// return the date of the last commit of the repository.
+    fn date_of_the_last_commit(&self) -> Option<String>;
```
vincenzopalazzo, line 40: This is a bad design, you should remove the method for the `Repository` Trait. It is an implementation-specific information, so this should stay in `coffee_github`

coffee_lib/src/types/mod.rs
```diff
@@ -109,6 +109,8 @@ pub mod response {
         pub local_name: String,
         pub url: String,
         pub plugins: Vec<Plugin>,
+        pub commit_id: String,
+        pub date: String,
```
vincenzopalazzo: These need to be optional types

coffee_storage/src/model/repository.rs
```diff
@@ -15,4 +15,6 @@ pub struct Repository {
     pub url: URL,
     pub plugins: Vec<Plugin>,
     pub branch: String,
+    pub git_head: Option<String>,
+    pub date_of_the_last_commit: Option<String>,
```
vincenzopalazzo, line 19: move `date_of_the_last_commit` tp `last_activity`

vincenzopalazzo (changes requested):
Needs some more works

coffee_core/src/coffee.rs
tareknaser: this is outdated
solved in 3b55f3e

coffee_lib/src/repository.rs
tareknaser: This is just the definition of these methods
The actual implementation is in `coffee_github/src/repository.rs`
We have to implement these methods here to be able to call them on variables on type `Repository` in `coffee_core/src/coffee.rs` 
For example in `list_remotes` method

vincenzopalazzo: >This is just the definition of these methods

This is the problem. You require this method in all the implementations of this trait, and this is wrong. You should implement this only inside `coffee_github`.

>We have to implement these methods here to be able to call them on variables on type Repository in coffee_core/src/coffee.rs
For example, in list_remotes method

Why not call just `get_info` in these places?

tareknaser: > Why not call just get_info in these places?
We definitely can

vincenzopalazzo: yeah so we should only expose the get_info and then implement in the `coffee_github` the others two methods

coffee_cmd/src/coffee_term/command_show.rs
```diff
@@ -54,15 +54,27 @@ pub fn show_remote_list(remote_list: Result<CoffeeRemote, CoffeeError>) -> Resul
...
 
     for repository in &repositories {
+        let commit_id = match &repository.commit_id {
+            Some(commit_id) => commit_id.to_owned(),
+            None => String::from(""),
+        };
+        let date = match &repository.date {
+            Some(date) => date.to_owned(),
+            None => String::from(""),
+        };
```
vincenzopalazzo, line 70: You can use also `repository.date.unwrap_or_default()`

coffee_core/src/coffee.rs
```diff
@@ -14,13 +14,14 @@ use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use tokio::process::Command;
 
+use chrono::{TimeZone, Utc};
```
vincenzopalazzo, line 17: this should be under the tokio import. We try to keep separate the our import and the library import.

coffee_core/src/coffee.rs
```diff
@@ -14,13 +14,14 @@ use serde::de::DeserializeOwned;
...
 use coffee_github::repository::Github;
 use coffee_lib::errors::CoffeeError;
 use coffee_lib::plugin_manager::PluginManager;
 use coffee_lib::repository::Repository;
 use coffee_lib::types::response::*;
 use coffee_lib::url::URL;
-use coffee_lib::{error, sh};
+use coffee_lib::{commit_id, error, get_repo_info, sh};
```
vincenzopalazzo, line 24: Now we should have the `use coffe_lib::prelude::*`

vincenzopalazzo (commented):
LGTM

I have some comments but this looks good we can add them later

Issue #194, opened 2023-08-13
use docker for default install


jackstar12:
Currently default plugin installation is only implemented for pip and poetry, which is suboptimal. 

Figuring this out for compiled languages would be pretty hard because you need to figure out how to detect / install each toolchain, which is why I propose using docker or podman (daemonless docker) for building the plugin into a container which cln can then run through a small generated script. Each language could have a default image or provide a custom one in `coffee.yaml`.

If docker / podman isn't installed on the machine we could still fallback to the current way of doing things.

I'd be willing to try implement this.

vincenzopalazzo:
>Currently default plugin installation is only implemented for pip and poetry, which is suboptimal.

Currently, with coffee.yml we support all the plugins, and I would like to see it as a standard because you can also have a missing python dir. but yeah you are right.

>If docker / podman isn't installed on the machine we could still fallback to the current way of doing things.

Yeah correct, do you think that this required some change inside the coffee core? or it is possible to do just with `coffee.yml`?

>I'd be willing to try implement this.

I would love to see it, please do if you need to chat more about it there is a Discussion open  in the github repository

Thanks for this good point

PR #195, opened 2023-08-26
fix(lib): simplify and improve sh! macro

jackstar12:
working on https://github.com/coffee-tools/coffee/issues/194 I noticed a issue with the `sh` macro when trying to execute a script that uses a pipe. Looking into the implementation I saw that it tries to manually parse out the commands, which is pretty error prone. so, instead of trying to manually parse the script and potentially making errors, let the actual `sh` command do the work correctly

### DIFF
--- a/coffee_lib/src/macros.rs
+++ b/coffee_lib/src/macros.rs
@@ -15,31 +15,27 @@ macro_rules! error {
 macro_rules! sh {
     ($root: expr, $script:expr, $verbose:expr) => {{
         let script = $script.trim();
-        let cmds = script.split("\n"); // Check if the script contains `\`
-        log::debug!("cmds: {:?}", cmds);
-        for cmd in cmds {
-            log::debug!("cmd {:?}", cmd);
-            let cmd_tok: Vec<&str> = cmd.split(" ").collect();
-            let command = cmd_tok.first().unwrap().to_string();
-            let mut cmd = Command::new(command);
-            cmd.args(&cmd_tok[1..cmd_tok.len()]);
-            cmd.current_dir($root);
-            let command = if $verbose {
-                cmd.spawn()
-                    .expect("Unable to run the command")
-                    .wait_with_output()
-                    .await?
-            } else {
-                cmd.output().await?
-            };
+        log::debug!("script: {:?}", script);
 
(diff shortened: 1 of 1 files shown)


coffee_lib/src/macros.rs
```diff
@@ -15,31 +15,27 @@ macro_rules! error {
...
+        log::debug!("script: {:?}", script);
 
-            if !command.status.success() {
-                let mut content = String::from_utf8(command.stderr).unwrap();
-                if content.trim().is_empty() {
-                    content = String::from_utf8(command.stdout).unwrap();
-                }
-                return Err(CoffeeError::new(2, &content));
+        let mut cmd = Command::new("sh");
+        cmd.args(&["-c", &script]);
```
vincenzopalazzo, line 21: Ah nice trick!

coffee_lib/src/macros.rs
```diff
@@ -15,31 +15,27 @@ macro_rules! error {
...
+        cmd.current_dir($root);
+
+        let command = if $verbose {
+            cmd.spawn()
+                .expect("Unable to run the command")
```
vincenzopalazzo, line 26:
```suggestion
                .map_err(|_| error!("Unable to run the command"))?
```

Does writing this code in this way works?

vincenzopalazzo (commented):
LGTM

I have just a comment on the `expect` call, we can return an error and avoid panicking? if this panic happens in the httpd can cause problem (maybe)

thanks! VERY NICE IMPROVEMENT

vincenzopalazzo:
@jackstar12 Are you still working on this?

jackstar12:
> @jackstar12 Are you still working on this? I would really love to merge went he review is addressed

yes, I was on vacation this week, I`ll get to it in the next few days

coffee_lib/src/macros.rs
jackstar12: yes

vincenzopalazzo (approved):
Thanks for working on this and fixing my bad sh macros!

ACK https://github.com/coffee-tools/coffee/pull/195/commits/785e8ac298c37ac945566b75a92815d37aabcd35

PR #197, opened 2023-09-19
Prepare Coffee Nurse Command Skeleton

tareknaser:


coffee_cmd/src/main.rs
```diff
@@ -123,10 +123,26 @@ async fn main() -> Result<(), CoffeeError> {
...
             Err(err) => Err(err),
         },
-        CoffeeCommand::Nurse {} => {
-            term::info!("Nurse command is not implemented");
-            Ok(())
-        }
+        CoffeeCommand::Nurse {} => match coffee.nurse().await {
+            Ok(val) => {
+                match val.status {
+                    NurseStatus::Sane => {
```
vincenzopalazzo, line 129: you can implement a `std::fmt::Display` on the NurseStatus and then call `to_string`. In this way you have a single call to `term::success`

coffee_core/src/nurse/chain.rs
```diff
@@ -48,16 +50,18 @@ pub struct RecoveryChainOfResponsibility {
...
         })
     }
 
-    pub async fn scan(&self) -> Result<(), CoffeeError> {
+    pub async fn scan(&self) -> Result<CoffeeNurse, CoffeeError> {
         for handler in self.handlers.iter() {
-            if let Some(strategy) = handler.clone().can_be_apply().await? {
-                strategy.patch().await?;
+            if let Some(strategy) = handler.clone().can_be_applied().await? {
+                return strategy.patch().await;
```
vincenzopalazzo, line 60:
```suggestion
                return strategy.patch().await;
```

not sure that this is good, a chan is a sequence of elements and you are returning to the first one that can be applied. 

But maybe the strategy that you have in mind to implement works in this way

coffee_core/src/nurse/chain.rs
```diff
@@ -48,16 +50,18 @@ pub struct RecoveryChainOfResponsibility {
 impl RecoveryChainOfResponsibility {
     pub async fn new() -> Result<Self, CoffeeError> {
         Ok(Self {
-            handlers: Vec::new(),
+            handlers: vec![Arc::new(GitRepositoryLocallyAbsentStrategy)],
         })
     }
```
vincenzopalazzo, line 55: The chanin usually takes the struct where you are making the operation on. 

If the chain of responsibility make an operation on the database only, you pass the database inside it as a pointer (Arc<..>).

If the chain of responsibility makes operation on a more general object like CoffeeManager, you pass down the coffee manager always in an `Arc` 

Arc without a Mutex inside is like a `const *void` pointer in C, A pointer that can not be modified so, in this case, the strategy should not modify the coffee manager, but just use it as an access point to information.

vincenzopalazzo (changes requested):

coffee_cmd/src/main.rs
tareknaser: Good point
will do

coffee_core/src/nurse/chain.rs
tareknaser: `can_be_applied` function should ideally serve as an indicator for applying a strategy if and only if the specific case it is designed to handle is encountered. 

I haven't really thought of the idea that we might have 2 or more cases at the same time.
We should be able to fix all the problems with running `coffee nurse` once 

Good catch!

coffee_core/src/nurse/chain.rs
tareknaser: The `handlers` should be a vector of the strategies to deal with corrupt states. right?

I guess the question is what to pass to `self.recovery_strategies.scan().await`  in `coffee.rs` file.
Initially, it appears that we may require mutable access to the CoffeeManager since certain scenarios may necessitate mutations.

For example, A repository in `CoffeeManager` is missing locally but it can't be restored using `git` for a reason or another. In this case, we remove it from the configuration.

coffee_core/src/nurse/chain.rs
vincenzopalazzo: >I haven't really thought of the idea that we might have 2 or more cases at the same time.

This is a common case for a chain of responsibility but looks like a good thing to experiment with and then return a vector of operation done by coffee.

So maybe the return should be a push_bach in the vector?

coffee_core/src/nurse/chain.rs
vincenzopalazzo: I see, so I will let you play with it. I will let you play with it

If you need to pass a mutable state I sure you need a `Arc<Mutex<CoffeeManager>>`

coffee_github/src/repository.rs
```diff
@@ -268,6 +268,46 @@ impl Repository for Github {
...
+                // retrieve the submodules
+                let submodules = repo.submodules().unwrap_or_default();
+                for (_, sub) in submodules.iter().enumerate() {
+                    let path =
+                        format!("{}/{}", &self.url.path_string, sub.path().to_str().unwrap());
+                    if let Err(err) = git2::Repository::clone(sub.url().unwrap(), &path) {
+                        return Err(error!("{}", err.message()));
+                    }
+                }
+
```
vincenzopalazzo, line 304: Nice, I was thinking if we may need to make a git helper to do this operation atomically


We need this code also during the `upgrade` command, but we can do this refactoring also later when you are working on the `upgrade`

coffee_lib/src/types/mod.rs
```diff
@@ -86,6 +86,7 @@ pub mod request {
 // Definition of the response types.
 pub mod response {
     use serde::{Deserialize, Serialize};
+    use std::fmt;
```
vincenzopalazzo, line 89: we can fix the import here?
```suggestion
    use std::fmt;
    
    use serde::{Deserialize, Serialize};
```

vincenzopalazzo (changes requested):
Really nice job. I really like this.

I wondered what happens if I delete the `repositories` directory and not just a sub-directory.

Maybe the directory is a default one and it is always created, right?

tareknaser:
> I wondered what happens if I delete the `repositories` directory and not just a sub-directory.
> 
> Maybe the directory is a default one and it is always created, right?

Yes, we call 
```rust
check_dir_or_make_if_missing(format!("{def_path}/repositories")).await?;
```
with each command

coffee_core/src/coffee.rs
```diff
@@ -442,8 +448,70 @@ impl PluginManager for CoffeeManager {
...
 
-    async fn nurse(&mut self) -> Result<(), CoffeeError> {
-        self.recovery_strategies.scan().await
+    async fn nurse(&mut self) -> Result<CoffeeNurse, CoffeeError> {
+        let status = self.recovery_strategies.scan(self).await?;
+        let mut nurse_actions: Vec<NurseStatus> = vec![];
+        for defect in status.defects.iter() {
+            log::debug!("defect: {:?}", defect);
+            match defect {
+                Defect::RepositoryLocallyAbsent(repos) => {
```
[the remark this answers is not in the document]
tareknaser: > are you checking the storage exist but the repositories directory no?

Yes
It's confusing now because it's the only case 
later we will also have `RepositoryAbsentInConfiguration`  which is the case when a repo exists on disk but not in `coffee` configuration

coffee_core/src/coffee.rs
```diff
@@ -442,8 +448,70 @@ impl PluginManager for CoffeeManager {
...
+        // There are 2 cases:
+        // 1. the repository can be recovered from the remote
+        // 2. the repository can't be recovered from the remote. In this case
+        //   we remove the repository from the coffee configuration.
+        for repo_name in repos.iter() {
+            // Get the repository from the name
+            let repo = self
+                .repos
+                .get_mut(repo_name)
+                .ok_or_else(|| error!("repository with name: {repo_name} not found"))?;
```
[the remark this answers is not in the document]
tareknaser: We get `repo_name` from the checker in `recovery_strategies` which checks the configuration and returns a vector of names of the repos missing but are actually in the configuration
and this line just retrieves the repo by its name so it really shouldn't ever fail

coffee_cmd/src/main.rs
```diff
@@ -123,10 +123,16 @@ async fn main() -> Result<(), CoffeeError> {
...
-        }
+        CoffeeCommand::Nurse {} => match coffee.nurse().await {
+            Ok(val) => {
+                // For every status in CoffeeNurse, print it
+                for status in val.status {
+                    term::success!("{}", status.to_string());
+                }
+                Ok(())
+            }
+            Err(err) => Err(err),
```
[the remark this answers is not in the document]
tareknaser: Yes, worth the effort
looks cleaner

coffee_core/src/coffee.rs
```diff
@@ -442,8 +448,70 @@ impl PluginManager for CoffeeManager {
...
+                    let mut actions = self.patch_repository_locally_absent(repos.to_vec()).await?;
+                    nurse_actions.append(&mut actions);
+                }
+            }
+        }
+
+        // If there was no actions taken by nurse, we return a sane status.
+        if nurse_actions.is_empty() {
+            nurse_actions.push(NurseStatus::Sane);
+        }
```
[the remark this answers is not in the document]
tareknaser: It's a tradeoff
easier to read but after a second thought, I guess it makes more sense to return an empty vector because we are now returning a vector of actions taken and an empty vector makes more sense in that context 
I will consider this in the next PR for `nurse`

coffee_github/src/repository.rs
tareknaser: the `upgrade` command needs to be refactored now after we changed the install path of plugins to a separate folder 
maybe I will also refactor the code here to be reused

coffee_cmd/src/coffee_term/command_show.rs
```diff
@@ -76,3 +76,51 @@ pub fn show_remote_list(remote_list: Result<CoffeeRemote, CoffeeError>) -> Resul
...
 }
+
+pub fn show_nurse_result(
+    nurse_result: Result<CoffeeNurse, CoffeeError>,
+) -> Result<(), CoffeeError> {
+    match nurse_result {
+        Ok(nurse) => {
+            // special case: if the nurse is sane
+            // we print a message and return
+            if nurse.status[0] == NurseStatus::Sane {
```
vincenzopalazzo, line 87: in the future we should like to have a struct

```rust
struct NurseStatus {
   defects: Vec<...>,
}

impl NurseStatus {
   fn is_sane(&self) -> bool {
        ....
    }
}
```

vincenzopalazzo (approved):
LGTM

Small comment but this can be a followup PR

vincenzopalazzo:
```
➜  coffee git:(pr197) cargo run --bin coffee -- --network testnet nurse
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/coffee --network testnet nurse`
╭────────────────────────────────────────────────╮
│ ●   Actions Taken        Affected repositories │
├────────────────────────────────────────────────┤
│ ●   Restored using Git   folgore-git           │
╰────────────────────────────────────────────────╯
```

Amazing!

PR #198, opened 2023-09-26
ci: update to the last cln version on master

vincenzopalazzo:
This is being added instead of the build-time setting for v23.11.


tareknaser (approved):
LGTM

PR #200, opened 2023-10-01
Add Command to List Plugins in a Remote Repository

tareknaser:
## Description
This pull request introduces a new command to `coffee`, allowing users to list plugins in a remote repository.
## Example:
To list plugins in a remote repository, simply run:
```bash
coffee remote <repository_name> --plugins
```
## Changes
- [x] Added a new HTTP endpoint to handle the command.
- [x] Implemented test cases to ensure the functionality is reliable.
- [x] Documented the new command and its usage.

### DIFF
--- a/coffee_cmd/src/main.rs
+++ b/coffee_cmd/src/main.rs
@@ -5,9 +5,10 @@ use clap::Parser;
 use radicle_term as term;
 
 use coffee_core::coffee::CoffeeManager;
+use coffee_lib::error;
 use coffee_lib::errors::CoffeeError;
 use coffee_lib::plugin_manager::PluginManager;
-use coffee_lib::types::response::UpgradeStatus;
+use coffee_lib::types::response::{CoffeeRemote, UpgradeStatus};
 
 use crate::cmd::CoffeeArgs;
 use crate::cmd::CoffeeCommand;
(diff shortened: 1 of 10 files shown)


vincenzopalazzo:
What happens if I do not specify `--plugins`?

tareknaser:
> What happens if I do not specify `--plugins`?

fails, `clap` handles that because of `requires` here 
```rust
#[arg(short, long, action = clap::ArgAction::SetTrue, requires = "remote-name")]
plugins: bool,
#[arg(
    requires = "plugins",
    name = "remote-name",
    help = "The name of the remote repository"
)]
name: Option<String>,
```


the output is 
```bash
% coffee remote lightningd          
error: the following required arguments were not provided:
  --plugins

Usage: coffee remote --plugins <remote-name>

For more information, try '--help'.
```

vincenzopalazzo:
>fails, clap handles that because of requires here

Without a flag, we can print the information about the repository.

We already do it inside the `coffee remote list`, maybe we can return the same result but with the same item?

coffee_cmd/src/main.rs
```diff
@@ -2,6 +2,7 @@ mod cmd;
 mod coffee_term;
 
 use clap::Parser;
+use coffee_lib::error;
```
vincenzopalazzo, line 5: move this down with the other coffee import

coffee_lib/src/plugin_manager.rs
```diff
@@ -36,6 +36,9 @@ pub trait PluginManager {
     /// list the remote repositories for the plugin manager.
     async fn list_remotes(&mut self) -> Result<CoffeeRemote, CoffeeError>;
 
+    /// List the plugins available in a remote repository.
+    async fn list_plugins_in_remote_repo(&self, name: &str) -> Result<CoffeeList, CoffeeError>;
```
vincenzopalazzo, line 40:
```suggestion
    async fn list_plugins_in_remote_repo(&self, name: &str) -> Result<CoffeeList, CoffeeError>;
```

due that this being a public method I would like to change the name to one a little bit more ergonomic like `remote_plugins` or `get_plugins_in_remote`.

`Remote repo` is kind confusing because the remote is a repository in the case of git

vincenzopalazzo (changes requested):
OK some very small suggestion change, but we are there.

Thanks

vincenzopalazzo (approved):
LGTM, thanks

PR #211, opened 2023-10-19
Enhance Error Message for Missing Core Lightning Configuration File

### LINKED ISSUE #209: coffee is not able to recover from cln missing
Let's imagine that I have my core lightning in an external SSD, and I do not have it plugged.

This is the log that I found inside my machine by running `RUST_LOG=debug coffee remote list`

It produces the following error

```
Error: CoffeeError { code: 1, msg: "No such file or directory (os error 2)" }
```

This is the log that it produces before returning the error

```
➜  cln-passip git:(master) RUST_LOG=debug coffee remote list
[2023-10-17T10:02:54Z INFO  coffee_core::config] creating coffee home at /home/vincent/.coffee
[2023-10-17T10:02:54Z DEBUG sled::pagecache::iterator] ordering before clearing tears: {0: 0}, max_header_stable_lsn: 0
[2023-10-17T10:02:54Z DEBUG sled::pagecache::iterator] in clean_tail_tears, found missing item in tail: None and we'll scan segments {0: 0} above lowest lsn 0
[2023-10-17T10:02:54Z DEBUG sled::pagecache::iterator] filtering out segments after detected tear at (lsn, lid) 22158
[2023-10-17T10:02:54Z DEBUG sled::pagecache::iterator] hit max_lsn 22158 in iterator, stopping
[2023-10-17T10:02:54Z DEBUG sled::pagecache::snapshot] zeroing the end of the recovered segment at lsn 0 between lids 22159 and 524287
[2023-10-17T10:02:54Z DEBUG sled::pagecache::blob_io] gc_blobs removing any blob with an lsn above 22159
[2023-10-17T10:02:54Z DEBUG sled::pagecache::segment] SA starting with tip 524288 stable -1 free {}
[2023-10-17T10:02:54Z DEBUG sled::pagecache::iobuf] starting log at recovered active offset 22159, recovered lsn 22159
[2023-10-17T10:02:54Z DEBUG sled::pagecache::iobuf] starting IoBufs with next_lsn: 22159 next_lid: 22159
[2023-10-17T10:02:54Z DEBUG sled::pagecache] load_snapshot loading pages from 0..4
[2023-10-17T10:02:54Z DEBUG coffee_core::coffee] repositories in store {"folgore-git": Repository { kind: Git, name: "folgore-git", url: URL { name: "folgore-git", url_string: "https://github.com/coffee-tools/folgore", path_string: "/home/vincent/.coffee/repositories/folgore-git", repo_name: "folgore.git" }, plugins: [Plugin { name: "folgore", root_path: "/home/vincent/.coffee/repositories/folgore-git", exec_path: "/home/vincent/.coffee/repositories/folgore-git/target/release/folgore_plugin", lang: Rust, conf: Some(Conf { plugin: Plugin { name: "folgore", version: "0.0.1", lang: "rust", deprecated: None, dependencies: None, install: Some("make coffee\n"), main: "target/release/folgore_plugin" } }), commit: Some("0f7a71b0d245cc3b180f8f4c35862df8cb799d87") }], branch: "main", git_head: Some("0f7a71b0d245cc3b180f8f4c35862df8cb799d87"), last_activity: Some("28/09/2023") }, "lnmetrics-git": Repository { kind: Git, name: "lnmetrics-git", url: URL { name: "lnmetrics-git", url_string: "https://github.com/LNOpenMetrics/go-lnmetrics.reporter", path_string: "/home/vincent/.coffee/repositories/lnmetrics-git", repo_name: "go-lnmetrics.reporter.git" }, plugins: [Plugin { name: "lnmetrics", root_path: "/home/vincent/.coffee/repositories/lnmetrics-git", exec_path: "/home/vincent/.coffee/repositories/lnmetrics-git/go-lnmetrics", lang: Go, conf: Some(Conf { plugin: Plugin { name: "lnmetrics", version: "0.0.6-rc2", lang: "go", deprecated: None, dependencies: None, install: Some("make dep\nmake coffee\n"), main: "go-lnmetrics" } }), commit: Some("5ad4c0fb9326950fc5887556d6cc0de73eb91d76") }], branch: "main", git_head: Some("5ad4c0fb9326950fc5887556d6cc0de73eb91d76"), last_activity: Some("12/06/2023") }}
[2023-10-17T10:02:54Z DEBUG coffee_core::nurse::strategy] Checking if repository folgore-git exists locally
[2023-10-17T10:02:54Z DEBUG coffee_core::nurse::strategy] Checking if repository lnmetrics-git exists locally
[2023-10-17T10:02:54Z DEBUG coffee_core::nurse::strategy] No repositories missing locally
[2023-10-17T10:02:54Z DEBUG coffee_core::coffee] Chain of responsibility status: ChainOfResponsibilityStatus { defects: [] }
[2023-10-17T10:02:54Z INFO  coffee_core::coffee] looking for the cln config: /run/media/vincent/VincentSSD/.lightning/bitcoin/config
[2023-10-17T10:02:54Z DEBUG sled::pagecache::logger] IoBufs dropped
```

@tareknaser Do you think that this is a good strategy to add inside the nurse command? We can suggest rerunning the lightning

tareknaser:
The error message displayed in case of a missing Core Lightning configuration file has been enhanced to provide users with more friendly and actionable information.

Now, when coffee fails to find the configuration file at the specified path, it advises users to resolve the issue by running the `coffee setup <cln_path>` command, helping them quickly address the problem and proceed smoothly.

Fixes #209

### DIFF
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -195,10 +195,12 @@ impl CoffeeManager {
         if self.config.cln_config_path.is_none() {
             return Ok(());
         }
-        let root = self.config.cln_root.clone().unwrap();
+        let root = self.config.cln_root.clone();
+        let root = root.ok_or_else(|| error!("cln root path not found."))?;
         let rpc = Client::new(format!("{root}/{}/lightning-rpc", self.config.network));
         self.rpc = Some(rpc);
-        let path = self.config.cln_config_path.clone().unwrap();
+        let path = self.config.cln_config_path.clone();
+        let path = path.ok_or_else(|| error!("cln config path not found."))?;
         let mut file = CLNConf::new(path.clone(), true);
         log::info!("looking for the cln config: {path}");
         file.parse()
@@ -217,7 +219,8 @@ impl CoffeeManager {
         self.config.cln_config_path = Some(path_with_network);
         self.config.cln_root = Some(cln_dir.to_owned());
         self.load_cln_conf().await?;
-        let mut conf = self.cln_config.clone().unwrap();
+        let conf = self.cln_config.clone();
+        let mut conf = conf.ok_or_else(|| error!("cln config path not found."))?;
         conf.add_subconf(self.coffee_cln_config.clone())
             .map_err(|err| error!("{}", &err.cause))?;
         conf.flush()?;
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -2,6 +2,7 @@
 
 use std::collections::HashMap;
 use std::fmt::Debug;
+use std::path::Path;
 use std::vec::Vec;
 use tokio::fs;
 
@@ -201,6 +202,17 @@ impl CoffeeManager {
         self.rpc = Some(rpc);
         let path = self.config.cln_config_path.clone();
         let path = path.ok_or_else(|| error!("cln config path not found."))?;
+
+        // The path is the path of the cln config file (eg. ~/.lightning/bitcoin/config)
+        // We need to check that the root folder (eg. ~/.lightning/bitcoin) exists
+        // otherwise we return an error.
+
+        // We can unwrap here because we are sure that the path ends with /config
+        let lightning_path = path.strip_suffix("/config").unwrap();
+        if !Path::new(&lightning_path).exists() {
+            return Err(error!("The path {lightning_path} does not exist"));
+        }
+
         let mut file = CLNConf::new(path.clone(), true);
         log::info!("looking for the cln config: {path}");
         file.parse()


coffee_core/src/coffee.rs
```diff
@@ -201,8 +201,17 @@ impl CoffeeManager {
...
+        file.parse().map_err(|err| {
+            CoffeeError::new(
+                err.core,
+                format!(
+                    "coffee was unable to locate Core Lightning configuration file at the specified path: {}.
+                     To resolve this, please run the following command: coffee setup <cln_path>.",
+                    path
+                )
+                .as_str(),
+            )
```
vincenzopalazzo, line 213: Mh I do think this is always right, if the file is malformed an error is returned, and in this way you are hiding it

vincenzopalazzo (changes requested):
If you want to go down this road, you should check if the path exists before create the CLNConf file.

coffee_core/src/coffee.rs
tareknaser: Do you prefer to keep the previous error and instead check if the path exists separately and return this coffee error if it's not found?

tareknaser: What is the solution that we should recommend to the user?

vincenzopalazzo: What I am saying that the parse function will not fails only if the file is missing, but will return an error also when the configuration file is not good

the parse function will fails if you put the following content

```
[bitcoin]
plugin=/path/bin
```

You return `coffee was unable to locate Core Lightning configuration file` but this is not true

tareknaser: > but will return an error also when the configuration file is not good

Ah yes I totally missed this case 
Thanks!

coffee_core/src/coffee.rs
```diff
@@ -195,10 +196,24 @@ impl CoffeeManager {
         if self.config.cln_config_path.is_none() {
             return Ok(());
         }
-        let root = self.config.cln_root.clone().unwrap();
+        let root = self
+            .config
+            .cln_root
+            .clone()
+            .ok_or_else(|| error!("cln root path not found, please run `coffee setup`"))?;
```
vincenzopalazzo, line 203:
```suggestion
        let root = self
            .config
            .cln_root;
       let root = root
            .ok_or_else(|| error!("cln root path not found, please run `coffee setup`"))?;
```

Avoid long chain, it is huggly

coffee_core/src/coffee.rs
```diff
@@ -195,10 +196,24 @@ impl CoffeeManager {
...
         let rpc = Client::new(format!("{root}/{}/lightning-rpc", self.config.network));
         self.rpc = Some(rpc);
-        let path = self.config.cln_config_path.clone().unwrap();
+        let path = self
+            .config
+            .cln_config_path
+            .clone()
+            .ok_or_else(|| error!("cln config path not found, please run `coffee setup`"))?;
```
vincenzopalazzo, line 210: same here

vincenzopalazzo (changes requested):

coffee_core/src/coffee.rs
tareknaser: Fixed

coffee_core/src/coffee.rs
```diff
@@ -195,10 +195,14 @@ impl CoffeeManager {
         if self.config.cln_config_path.is_none() {
             return Ok(());
         }
-        let root = self.config.cln_root.clone().unwrap();
+        let root = self.config.cln_root.clone();
+        let root =
+            root.ok_or_else(|| error!("cln root path not found, please run `coffee setup`"))?;
         let rpc = Client::new(format!("{root}/{}/lightning-rpc", self.config.network));
```
vincenzopalazzo, line 201: I do not understand this, this will never fails.

Where you are checking if `cln_root` exist on the disk?

vincenzopalazzo (changes requested):
I think this is wrong

coffee_core/src/coffee.rs
```diff
@@ -195,10 +195,12 @@ impl CoffeeManager {
...
             return Ok(());
         }
-        let root = self.config.cln_root.clone().unwrap();
+        let root = self.config.cln_root.clone();
+        let root = root.ok_or_else(|| error!("cln root path not found."))?;
         let rpc = Client::new(format!("{root}/{}/lightning-rpc", self.config.network));
         self.rpc = Some(rpc);
-        let path = self.config.cln_config_path.clone().unwrap();
+        let path = self.config.cln_config_path.clone();
+        let path = path.ok_or_else(|| error!("cln config path not found."))?;
```
vincenzopalazzo, line 203: what about the path is `Some` but do not exist?

vincenzopalazzo (changes requested):

coffee_core/src/coffee.rs
tareknaser: The next line is 
```rust
let mut file = CLNConf::new(path.clone(), true);
```
where we set `create_if_missing` to `true` 
This is what happens when the user first runs coffee

vincenzopalazzo: Mh I do not think this is right, if the `path` do not exist this code do not make sense because `CLNConf::new(path.clone(), true);` is not able to find the path where to create the file.

Please test the user case that I described https://github.com/coffee-tools/coffee/issues/209

tareknaser: You are right.
This line will handle the case that the root path exists but the configuration file doesn't.
However, it will fail if the root path doesn't exist.

tareknaser: Added another check for this case.

vincenzopalazzo: >This line will handle the case that the root path exists but the configuration file doesn't.

This is done already by CLNConf

tareknaser: I am not sure `CLNConf` handles this case.
Could you give it another try?
To reproduce:
1- delete ` ~/.lightning/bitcoin` directory
2- try running coffee commands (eg. `coffee remote list`)
You will get the same error mentioned in the original issue

What `CLNConf` handles is the case when ` ~/.lightning/bitcoin/config` is missing but ` ~/.lightning/bitcoin` exists.

vincenzopalazzo: >I am not sure CLNConf handles this case.

yes is does, you need to do `CLNConf::new(path.clone(), false);` if the file do not exist it return an error

>What CLNConf handles is the case when  ~/.lightning/bitcoin/config is missing but  ~/.lightning/bitcoin exists.

Correct, you should check if the `path` exist

coffee_core/src/coffee.rs
```diff
@@ -201,6 +202,17 @@ impl CoffeeManager {
         self.rpc = Some(rpc);
         let path = self.config.cln_config_path.clone();
         let path = path.ok_or_else(|| error!("cln config path not found."))?;
+
+        // The path is the path of the cln config file (eg. ~/.lightning/bitcoin/config)
+        // We need to check that the root folder (eg. ~/.lightning/bitcoin) exists
+        // otherwise we return an error.
+
+        // We can unwrap here because we are sure that the path ends with /config
+        let lightning_path = path.strip_suffix("/config").unwrap();
```
vincenzopalazzo, line 211: If you need to strip this, this means that you are checking in the wrong place.

The check need to be done before we build the path with the `/config`

coffee_core/src/coffee.rs
```diff
@@ -201,6 +202,17 @@ impl CoffeeManager {
         self.rpc = Some(rpc);
         let path = self.config.cln_config_path.clone();
         let path = path.ok_or_else(|| error!("cln config path not found."))?;
+
+        // The path is the path of the cln config file (eg. ~/.lightning/bitcoin/config)
+        // We need to check that the root folder (eg. ~/.lightning/bitcoin) exists
+        // otherwise we return an error.
+
+        // We can unwrap here because we are sure that the path ends with /config
```
vincenzopalazzo, line 210:
```
// SAFETY: We can unwrap here because we are sure that the path ends with /config
```

vincenzopalazzo (changes requested):

vincenzopalazzo:
I think this code should be simpler, and I try to see if this is true https://github.com/coffee-tools/coffee/pull/220

tareknaser:
Closing this pull request and going ahead with #220. It has the same fix but is simpler.

PR #220, opened 2023-12-30
fix(cmd): check if the ln path that we know are good

vincenzopalazzo:
Sometimes we may do operations like delete a directory that is linked with coffee, but we forget it.

So this code makes sure that we are operating on a piece of safe information.

Fixes https://github.com/coffee-tools/coffee/issues/209

Alternative to https://github.com/coffee-tools/coffee/pull/211 this should fix the real problem


tareknaser (approved):
Thanks for the simple fix 
I tried it and it works as expected

PR #230, opened 2024-02-06
Only Upgrade Plugins if Needed

tareknaser:
Revise the plugin upgrade logic to only perform upgrades to plugins if the repository was updated. This prevents unnecessary plugin upgrades.

### DIFF
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -352,10 +352,18 @@ impl PluginManager for CoffeeManager {
             .ok_or_else(|| error!("Repository with name: `{}` not found", repo))?;
 
         let status = repository.upgrade(&self.config.plugins, verbose).await?;
-        for plugins in status.plugins_effected.iter() {
-            self.remove(plugins).await?;
-            self.install(plugins, verbose, false).await?;
+
+        // if status is not up to date, we need to update the plugins as well
+        match status.status {
+            UpgradeStatus::Updated(_, _) => {
+                for plugins in status.plugins_effected.iter() {
+                    self.remove(plugins).await?;
+                    self.install(plugins, verbose, false).await?;
+                }
+            }
+            _ => {}
         }
+
         self.flush().await?;
         Ok(status)
     }


coffee_core/src/coffee.rs
```diff
@@ -352,10 +352,18 @@ impl PluginManager for CoffeeManager {
...
+
+        // if status is not up to date, we need to update the plugins as well
+        match status.status {
+            UpgradeStatus::UpToDate(_, _) => {}
+            _ => {
+                for plugins in status.plugins_effected.iter() {
+                    self.remove(plugins).await?;
+                    self.install(plugins, verbose, false).await?;
+                }
+            }
```
vincenzopalazzo, line 364: we can do the oppisite? `UpdateStatus::Upgraded`
```suggestion
            UpdateStatus::Upgraded(_, _) => {
                for plugins in status.plugins_effected.iter() {
                    self.remove(plugins).await?;
                    self.install(plugins, verbose, false).await?;
                }
            },
            _ => {},
```

This is better to avoid bug, if we add a new state we know that only when the plugin is updated we run the remove and install steps

coffee_core/src/coffee.rs
```diff
@@ -352,10 +352,18 @@ impl PluginManager for CoffeeManager {
...
-            self.remove(plugins).await?;
-            self.install(plugins, verbose, false).await?;
+
+        // if status is not up to date, we need to update the plugins as well
+        match status.status {
+            UpgradeStatus::UpToDate(_, _) => {}
+            _ => {
+                for plugins in status.plugins_effected.iter() {
+                    self.remove(plugins).await?;
+                    self.install(plugins, verbose, false).await?;
```
vincenzopalazzo, line 362:
```suggestion
                    self.install(plugins, verbose, verbose).await?;
```

We must pass down the verbose field

vincenzopalazzo (changes requested):

coffee_core/src/coffee.rs
tareknaser: I thought about this as well but figured we only wouldn't want to upgrade the plugins if the repository is up to date. 
but I don't have a strong opinion on this one

coffee_core/src/coffee.rs
tareknaser: This is `try_dynamic` 
`verbose` is already passed

coffee_core/src/coffee.rs
vincenzopalazzo: it is better never fail to `_` case, and if we need it is better fall in the case where we do nothing

coffee_core/src/coffee.rs
vincenzopalazzo: mh nice, we may want also thing about `try_dynamic`

Issue #234, opened 2024-02-07
we should move the repository directory inside a network directory


vincenzopalazzo:
I try to run 

```
➜  ~ coffee --network testnet remote add folgore-git https://github.com/coffee-tools/folgore.git
✓ Remote added!
➜  ~ coffee remote add folgore-git https://github.com/coffee-tools/folgore.git
✗ Fetch remote from https://github.com/coffee-tools/folgore.git error: Error while add remote: code: 1, msg: repository with name: folgore-git already exists
✗ Error: code: 1, msg: repository with name: folgore-git already exists
```

This must not happen, because for each network we can add a different remote (e.g: a fork for testnet) so we should move the `repository` inside the `network` subdirectory

tareknaser:
I am fully with this.

There are so many scenarios where this feature would come in handy. This would also be useful when we add `coffee upgrade --branch <branch_name>`

vincenzopalazzo:
Yeah this si my fault that I tough that the repository was not needed as a network information

PR #239, opened 2024-02-07
feat(core): make a repository dir for each network

vincenzopalazzo:
This is the more sane thing to do because each network can have a different repository version.

Link: https://github.com/coffee-tools/coffee/issues/234

### DIFF
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -383,7 +383,8 @@ impl PluginManager for CoffeeManager {
         if self.repos.contains_key(name) {
             return Err(error!("repository with name: {name} already exists"));
         }
-        let url = URL::new(&self.config.root_path, url, name);
+        let local_path = format!("{}/{}", self.config.root_path, self.config.network);
+        let url = URL::new(&local_path, url, name);
         log::debug!("remote adding: {} {}", name, &url.url_string);
         let mut repo = Github::new(name, &url);
         repo.init().await?;
--- a/coffee_core/src/config.rs
+++ b/coffee_core/src/config.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 use std::env;
 
 use crate::CoffeeOperation;
-use coffee_lib::utils::check_dir_or_make_if_missing;
+use coffee_lib::utils::{check_dir_or_make_if_missing, move_dir_if_exist};
 use coffee_lib::{errors::CoffeeError, plugin::Plugin};
 
 use crate::CoffeeArgs;
@@ -65,7 +65,9 @@ impl CoffeeConf {
 
         check_dir_or_make_if_missing(format!("{def_path}/{}", coffee.network)).await?;
         check_dir_or_make_if_missing(format!("{def_path}/{}/plugins", coffee.network)).await?;
-        check_dir_or_make_if_missing(format!("{def_path}/repositories")).await?;
+        let repo_dir = format!("{def_path}/{}/repositories", coffee.network);
+        move_dir_if_exist(&format!("{def_path}/repositories"), &repo_dir).await?;
+        check_dir_or_make_if_missing(repo_dir).await?;
         // after we know all the information regarding
         // the configuration we try to see if there is
         // something stored already to the disk.
--- a/coffee_lib/src/utils.rs
+++ b/coffee_lib/src/utils.rs
@@ -2,6 +2,7 @@ use super::macros::error;
 use std::path::Path;
 
 use tokio::fs::create_dir;
+use tokio::fs::rename;
 
 use crate::errors::CoffeeError;
 
@@ -24,6 +25,14 @@ pub async fn check_dir_or_make_if_missing(path: String) -> Result<(), CoffeeErro
     if !Path::exists(Path::new(&path.to_owned())) {
         create_dir(path.clone()).await?;
         log::debug!("created dir {path}");
+    }
+    Ok(())
+}
+
+pub async fn move_dir_if_exist(origin: &str, destination: &str) -> Result<(), CoffeeError> {
+    if Path::exists(Path::new(&origin)) {
+        rename(origin, destination).await?;
+        log::debug!("move dir from {origin} to {destination}");
     }
     Ok(())
 }
--- a/git-bugreport-2024-02-07-1353.txt
+++ b/git-bugreport-2024-02-07-1353.txt
@@ -1,0 +1,32 @@
+Thank you for filling out a Git bug report!
+Please answer the following questions to help us understand your issue.
+
+What did you do before the bug happened? (Steps to reproduce your issue)
+
+What did you expect to happen? (Expected behavior)
+
+What happened instead? (Actual behavior)
+
+What's different between what you expected and what actually happened?
+
+Anything else you want to add:
+
+Please review the rest of the bug report below.
+You can delete any lines you don't wish to share.
+
+
+[System Info]
+git version:
+git version 2.43.0
+cpu: x86_64
+no commit associated with this build
+sizeof-long: 8
+sizeof-size_t: 8
+shell-path: /bin/sh
+uname: Linux 6.6.15-2-lts #1 SMP PREEMPT_DYNAMIC Fri, 02 Feb 2024 17:04:24 +0000 x86_64
+compiler info: gnuc: 13.2
+libc info: glibc: 2.39
+$SHELL (typically, interactive shell): /usr/bin/zsh
+
+
+[Enabled Hooks]


coffee_lib/src/types/mod.rs
```diff
@@ -138,6 +138,7 @@ pub mod response {
         // A patch operation when a git repository is present in the coffee configuration
         // but is absent from the local storage.
         RepositoryLocallyAbsent(Vec<String>),
+        CoffeeGlobalrepoCleanup(Vec<(String, String)>),
```
tareknaser, line 141: It's not clear what `Vec<(String, String)>` represents
we can add a doc comment here
```suggestion
        /// Affected network, path
        CoffeeGlobalrepoCleanup(Vec<(String, String)>),
```

coffee_core/src/coffee.rs
```diff
@@ -496,6 +499,16 @@ impl PluginManager for CoffeeManager {
                     let mut actions = self.patch_repository_locally_absent(repos.to_vec()).await?;
                     nurse_actions.append(&mut actions);
                 }
+                Defect::CoffeeGlobalrepoCleanup(networks) => {
+                    for (network, path) in networks {
+                        let global_repo = format!("{}/repositories", self.config.root_path);
+                        if !Path::exists(&Path::new(&path)) {
```
tareknaser, line 505: A clippy warning
```suggestion
                        if !Path::exists(Path::new(&path)) {
```

coffee_core/src/coffee.rs
```diff
@@ -496,6 +499,16 @@ impl PluginManager for CoffeeManager {
...
+                            move_dir_if_exist(&global_repo, &path).await?;
```
tareknaser, line 506:
```suggestion
                            move_dir_if_exist(&global_repo, path).await?;
```

coffee_core/src/nurse/strategy.rs
```diff
@@ -74,3 +74,28 @@ impl Handler for GitRepositoryLocallyAbsentStrategy {
...
+        self: Arc<Self>,
+        coffee: &CoffeeManager,
+    ) -> Result<Option<Defect>, CoffeeError> {
+        let networks = ["testnet", "signet", "bitcoin", "liquid"];
+        // check if the repository subdirectory has the repository directory
+        // inside the subdirectory.
+        let mut directory_moving = vec![];
+        for network in networks {
+            let subpath_repo = format!("{}/{network}/repositories", coffee.config.root_path);
+            if !Path::exists(&Path::new(&subpath_repo)) {
```
tareknaser, line 92:
```suggestion
            if !Path::exists(Path::new(&subpath_repo)) {
```

coffee_cmd/src/coffee_term/command_show.rs
```diff
@@ -102,10 +102,14 @@ pub fn show_nurse_result(
                     NurseStatus::RepositoryLocallyRemoved(_) => {
                         "Removed from local storage".to_string()
                     }
+                    NurseStatus::MovingGlobalRepostoryTo(_) => {
+                        format!("Moving Global repository directory")
```
tareknaser, line 106:
```suggestion 
                        "Moving Global repository directory".to_string()
```

coffee_core/src/coffee.rs
tareknaser, line 506: here we are iterating over the networks to see if the repositories sub-directory exists and if it doesn't we move the global repository inside it.

but this way, in the second iteration, there is no `global_repo` and `move_dir_if_exist` will do nothing

tareknaser (changes requested):
I also have a concern. The code in this PR handles moving the global repository to be network specific but does it handle the information in coffee configuration as well?

So, if we run `coffee nurse` and then see `coffee remote list` for a network. is the output sane?

coffee_core/src/coffee.rs
vincenzopalazzo: Good catch, we should copy it and remove the global repository at the end of the process, good catch

vincenzopalazzo:
ok I pushed the review you suggested @tareknaser tomorrow I will take a look if this PR is completed, but I guess yes?

I m not sure just when run the nurse command

tareknaser:
just a reminder that https://github.com/coffee-tools/coffee/pull/239#pullrequestreview-1873635869 is still not resolved

vincenzopalazzo:
Ok I think we must complete this, I ran into some issues that are really embracing

>So, if we run coffee nurse and then see the coffee remote list for a network. is the output sane?

Yes, why they should not be sane @tareknaser? I am missing something around your point?

vincenzopalazzo:
ok fixed a couple of bugs found with the nurse command.

Now the nurse command will migrate the directory, and then restore the repositories to have a fresh copy of them 😄 

Review are welcome

vincenzopalazzo:
With the current code I had to run two times the nurse command to have all fixed

```
➜  ~ coffee nurse
╭────────────────────────────────────────────────────────────────╮
│ ●   Actions Taken                        Affected repositories │
├────────────────────────────────────────────────────────────────┤
│ ●   Moving Global repository directory   testnet               │
│ ●   Moving Global repository directory   bitcoin               │
╰────────────────────────────────────────────────────────────────╯
➜  ~ coffee remote list
Error: CoffeeError { code: 1, msg: "Coffee found some defects in the configuration. Please run `coffee nurse` to fix them.\n                    If you are want to skip the verification, please add the `--skip-verify ` flag to the command." }
➜  ~ coffee nurse
╭─────────────────────────────────────────────────────╮
│ ●   Actions Taken        Affected repositories      │
├─────────────────────────────────────────────────────┤
│ ●   Restored using Git   summars, teos, folgore-git │
╰─────────────────────────────────────────────────────╯
➜  ~ coffee remote list
● List of repositories
╭───────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ●   Repository Alias   URL                                            N. Plugins   Git HEAD   Last Update │
├───────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ ●   folgore-git        https://github.com/coffee-tools/folgore        1            aedd703    12/03/2024  │
│ ●   summars            https://github.com/daywalker90/summars         1            c73633a    18/03/2024  │
│ ●   teos               https://github.com/vincenzopalazzo/rust-teos   1            507b8e2    09/04/2024  │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Maybe this is a way to do stuff but not sure if it is so good from implementation view point

vincenzopalazzo:
It breaks the upgrade command

```
➜  ~ coffee upgrade folgore-git -v
HEAD is now at 09189fb deps: changing the git url for esplora
cp: cannot stat '/home/vincent/.coffee/repositories/folgore-git': No such file or directory
✗ Error: code: 2, msg:
```

coffee_core/src/coffee.rs
```diff
@@ -545,12 +544,27 @@ impl PluginManager for CoffeeManager {
                     let mut actions = self.patch_repository_locally_absent(repos.to_vec()).await?;
                     nurse_actions.append(&mut actions);
                 }
+                Defect::CoffeeGlobalrepoCleanup(networks) => {
+                    let global_repo = format!("{}/repositories", self.config.root_path);
+                    for (network, path) in networks {
+                        log::info!("{network} - {path}");
+                        check_dir_or_make_if_missing(path.to_owned()).await?;
+                        if !Path::exists(Path::new(&path)) {
+                            copy_dir_if_exist(&global_repo, path).await?;
```
tareknaser, line 553: `copy_dir_if_exist` doesn't return an error if the origin folder doesn't exist so if the global repositories folder doesn't exist, we wouldn't know.

Is this intended?

coffee_core/src/nurse/strategy.rs
```diff
@@ -74,3 +74,39 @@ impl Handler for GitRepositoryLocallyAbsentStrategy {
...
 }
+
+/// Stategy for migration of the repository global directory
+/// to a local directory for each network, see [1]
+///
+/// This is a strategy tht return the list of network that
+/// needs to be migrated to to the new repository configuration.
+///
+/// [1] https://github.com/coffee-tools/coffee/issues/234
+pub struct CoffeeRepositoryDirCleanUp;
```
tareknaser, line 85: Can we rename it to `CoffeeRepositoryDirCleanUpStrategy` to keep the pattern?

coffee_core/src/nurse/strategy.rs
```diff
@@ -74,3 +74,39 @@ impl Handler for GitRepositoryLocallyAbsentStrategy {
...
     }
 }
+
+/// Stategy for migration of the repository global directory
+/// to a local directory for each network, see [1]
+///
+/// This is a strategy tht return the list of network that
+/// needs to be migrated to to the new repository configuration.
+///
+/// [1] https://github.com/coffee-tools/coffee/issues/234
```
tareknaser, line 84: To make the link valid
```suggestion
/// Stategy for migration of the repository global directory
/// to a local directory for each network. See [related issue]
///
/// This is a strategy that returns the list of networks that
/// need to be migrated to to the new repository configuration.
///
/// [issue]: https://github.com/coffee-tools/coffee/issues/234
```

coffee_lib/src/types/mod.rs
```diff
@@ -150,6 +150,8 @@ pub mod response {
         // A patch operation when a git repository is present in the coffee configuration
         // but is absent from the local storage.
         RepositoryLocallyAbsent(Vec<String>),
+        /// (Affected network, path)
+        CoffeeGlobalrepoCleanup(Vec<(String, String)>),
```
tareknaser, line 154: We could do just `Vec<String>` representing affected networks

coffee_core/src/nurse/strategy.rs
```diff
@@ -74,3 +74,39 @@ impl Handler for GitRepositoryLocallyAbsentStrategy {
...
+        let mut directory_moving = vec![];
+        for network in networks {
+            let network_dir = format!("{}/{network}", coffee.config.root_path);
+            if !Path::exists(Path::new(&network_dir)) {
+                log::debug!("network dir `{network_dir}` not found");
+                continue;
+            }
+            let subpath_repo = format!("{}/{network}/repositories", coffee.config.root_path);
+            if !Path::exists(Path::new(&subpath_repo)) {
+                directory_moving.push((network.to_string(), subpath_repo));
```
tareknaser, line 104: If we know the network, we could easily build the repositories path. 
What do you think?

coffee_lib/src/types/mod.rs
tareknaser, line 154:
```suggestion
        CoffeeGlobalRepoCleanup(Vec<(String, String)>),
```

coffee_cmd/src/coffee_term/command_show.rs
```diff
@@ -86,6 +89,52 @@ pub fn show_remote_list(remote_list: Result<CoffeeRemote, CoffeeError>) -> Resul
...
+                            .map(|(network, _)| format!("{network} "))
+                            .collect::<String>()
+                            .trim_end()
+                    );
+                    table.push([
+                        term::format::positive("●").into(),
+                        term::format::bold(defect.to_owned()),
+                        term::format::highlight("_".to_owned()),
+                    ]);
+                }
```
tareknaser, line 130:
```suggestion
                Defect::CoffeeGlobalrepoCleanup(networks) => {
                    let defect = "Network specific repository missing";
                    let networks = networks
                        .iter()
                        .map(|(network, _)| network.clone())
                        .collect::<Vec<String>>();
                    table.push([
                        term::format::positive("●").into(),
                        term::format::bold(defect.to_owned()),
                        term::format::highlight(networks.join(", ")),
                    ]);
                }
```

What do you think?

coffee_core/src/coffee.rs
```diff
@@ -573,6 +587,7 @@ impl PluginManager for CoffeeManager {
                 .get_mut(repo_name)
                 .ok_or_else(|| error!("repository with name: {repo_name} not found"))?;
 
+            repo.change_root_path(&self.config.path());
```
tareknaser, line 590: When we change the root path of the repository, are plugins' path also modified to reflect this change?

tareknaser (commented):
Thanks for putting this together. I like `coffee nurse --verify` output

I think the method to handle this nurse strategy should handle all the related logic so we don't need to modify `patch_repository_locally_absent` for example

coffee_core/src/coffee.rs
```diff
@@ -545,12 +544,27 @@ impl PluginManager for CoffeeManager {
...
+                    for (network, path) in networks {
+                        log::info!("{network} - {path}");
+                        check_dir_or_make_if_missing(path.to_owned()).await?;
+                        if !Path::exists(Path::new(&path)) {
+                            copy_dir_if_exist(&global_repo, path).await?;
+                        }
+                        nurse_actions
+                            .push(NurseStatus::MovingGlobalRepostoryTo(network.to_owned()));
+                    }
+                    rm_dir_if_exist(&global_repo).await?;
```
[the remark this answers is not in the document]
vincenzopalazzo: Agreee

coffee_core/src/coffee.rs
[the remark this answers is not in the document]
vincenzopalazzo: Thanks!

coffee_core/src/nurse/strategy.rs
vincenzopalazzo: What is the purpose here to build the repositories path? I am missing something sorry

coffee_cmd/src/coffee_term/command_show.rs
vincenzopalazzo: thanks!

coffee_core/src/coffee.rs
vincenzopalazzo: I think no, and we sould too right?

tareknaser: Yes we should

We have 2 options to handle this change (moving repositories folder from being global to being network specific including plugins' paths information):
- Handle everything in the nurse strategy patch method
- Implement a method to `repository` that handles it (similar to `change_root_path` in this PR)

I prefer the first option since I don't really see us using this method anywhere else in the code

coffee_core/src/nurse/strategy.rs
tareknaser: This comment is a follow up for https://github.com/coffee-tools/coffee/pull/239#discussion_r1558110379

I mean if we know the affected networks, we can build the repositories directory path specific to this network so no need to store it 
```rust
/// (Affected network, path)
CoffeeGlobalrepoCleanup(Vec<(String, String)>),
```

coffee_core/src/coffee.rs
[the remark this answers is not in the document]
vincenzopalazzo: Changin the approach I am reindexing the repository from crash, due that we are keeping the state with git, we can safely do it it.

The only missing point is that we are losing the commit where we are closing the repository, for the moment it is fine

coffee_core/src/nurse/strategy.rs
vincenzopalazzo: Agree thanks

coffee_core/src/nurse/strategy.rs
vincenzopalazzo: Ah, debugging porpuse for the moment, but I agree this code can be optimize

coffee_core/src/coffee.rs
vincenzopalazzo: just reindexing the repository with `add_remote` did the job in a smart way

coffee_cmd/src/coffee_term/command_show.rs
```diff
@@ -86,6 +89,49 @@ pub fn show_remote_list(remote_list: Result<CoffeeRemote, CoffeeError>) -> Resul
...
 
+pub fn show_nurse_verify(nurse_verify: &ChainOfResponsibilityStatus) -> Result<(), CoffeeError> {
+    if nurse_verify.defects.is_empty() {
+        term::success!("Coffee configuration is not corrupt! No need to run coffee nurse");
+    } else {
+        let mut table = radicle_term::Table::new(TableOptions::bordered());
+        table.push([
+            term::format::dim(String::from("●")),
+            term::format::bold(String::from("Defects")),
+            term::format::bold(String::from("Affected repositories")),
```
tareknaser, line 100: In case the defect is `CoffeeGlobalRepoCleanup`, this column will be affected networks (not repositories) 
Maybe we could keep the message generic 
```suggestion
            term::format::bold(String::from("Affected")),
```

coffee_core/src/nurse/strategy.rs
```diff
@@ -74,3 +74,32 @@ impl Handler for GitRepositoryLocallyAbsentStrategy {
...
     }
 }
+
+/// Stategy for migration of the repository global directory
+/// to a local directory for each network. See [related issue]
+///
+/// This is a strategy that returns the list of networks that
+/// need to be migrated to to the new repository configuration.
+///
+/// [issue]: https://github.com/coffee-tools/coffee/issues/234
```
tareknaser, line 84: I think I mistyped it in my first suggestion 😃
```suggestion
/// [related issue]: https://github.com/coffee-tools/coffee/issues/234
```

tareknaser (commented):
Thanks. This resolves the initial problem, but I believe there might still be a lurking bug.

tareknaser:
Currently, remote repositories are still global in the coffee configuration, not specific to individual networks. To reproduce:
```bash
coffee remote add lightningd https://github.com/lightningd/plugins.git
```
```bash
coffee -n testnet remote add folgore https://github.com/coffee-tools/folgore.git 
```
```bash
coffee remote list 
```
```bash
coffee -n testnet remote list
```
```bash
coffee install folgore # Shouldn't work
```
```bash
coffee remote rm folgore
```
```bash
coffee -n testnet remote list
```
## Expected
Output for `coffee remote list`: `lightningd`
Output for `coffee -n testnet remote list`: `folgore`
## Output
```bash
tareknasser@Tareks-MacBook ~ % coffee remote add lightningd https://github.com/lightningd/plugins.git
✓ Remote added!
tareknasser@Tareks-MacBook ~ % coffee -n testnet remote add folgore https://github.com/coffee-tools/folgore.git 
✓ Remote added!
tareknasser@Tareks-MacBook ~ % coffee remote list
● List of repositories
╭──────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ●   Repository Alias   URL                                       N. Plugins   Git HEAD   Last Update │
├──────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ ●   folgore            https://github.com/coffee-tools/folgore   1            09189fb    20/03/2024  │
│ ●   lightningd         https://github.com/lightningd/plugins     19           fef7ded    12/04/2024  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────╯
tareknasser@Tareks-MacBook ~ % coffee -n testnet remote list
● List of repositories
╭──────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ●   Repository Alias   URL                                       N. Plugins   Git HEAD   Last Update │
├──────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ ●   lightningd         https://github.com/lightningd/plugins     19           fef7ded    12/04/2024  │
│ ●   folgore            https://github.com/coffee-tools/folgore   1            09189fb    20/03/2024  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────╯
tareknasser@Tareks-MacBook ~ % coffee install folgore                 
✓ Compiling and installing
✓ Plugin folgore Compiled and Installed
tareknasser@Tareks-MacBook ~ % coffee remote rm folgore
✓ Remote removed!
tareknasser@Tareks-MacBook ~ % coffee -n testnet remote list
● List of repositories
╭────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ●   Repository Alias   URL                                     N. Plugins   Git HEAD   Last Update │
├────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ ●   lightningd         https://github.com/lightningd/plugins   19           fef7ded    12/04/2024  │
╰────────────────────────────────────────────────────────────────────────────────────────────────────╯

```

coffee_core/src/coffee.rs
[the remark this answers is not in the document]
tareknaser: > The only missing point is that we are losing the commit where we are closing the repository, for the moment it is fine

I agree. This is a clean solution

coffee_cmd/src/coffee_term/command_show.rs
vincenzopalazzo: I was thinking to move this label to action, what do you think?

vincenzopalazzo:
I think your case can be written in some integration testing that we have, so looking in doing it

vincenzopalazzo:
Thanks for this amazing catch @tareknaser 

I decided to fix our tests and write your example case in our test suite, to run it, run just `make integration ARGS="-- --test-threads=4 test_migrated_repository_to_local_one"`

P.S: pushed the test in a single single commit, and I find also a bug inside the nurse strategy that was not checking if there was the root repositories dir.

So now we should be ok

coffee_core/src/nurse/strategy.rs
```diff
@@ -74,3 +74,37 @@ impl Handler for GitRepositoryLocallyAbsentStrategy {
...
+///
+/// This is a strategy that returns the list of networks that
+/// need to be migrated to to the new repository configuration.
+///
+/// [related issue]: https://github.com/coffee-tools/coffee/issues/234
+pub struct CoffeeRepositoryDirCleanUpStrategy;
+
+#[async_trait]
+impl Handler for CoffeeRepositoryDirCleanUpStrategy {
+    async fn can_be_applied(
```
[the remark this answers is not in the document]
vincenzopalazzo: You are right!

tests/src/coffee_integration_tests.rs
```diff
@@ -700,3 +724,87 @@ pub async fn test_nurse_repository_missing_on_disk() {
...
+        .add_remote(
+            "lightningd",
+            "https://github.com/lightningd/plugins.git",
+            false,
+        )
+        .await
+        .unwrap();
+
+    let test_remotes = coffee_testnet.list_remotes().await.unwrap();
+    std::mem::drop(coffee_testnet_m);
```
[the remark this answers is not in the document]
vincenzopalazzo: Because we use the same database

vincenzopalazzo:
Find a bug when I try to add two times the same repo but in different network

```
➜  ~ coffee remote list
● List of repositories
╭──────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ●   Repository Alias   URL                                       N. Plugins   Git HEAD   Last Update │
├──────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ ●   folgore-git        https://github.com/coffee-tools/folgore   1            09189fb    20/03/2024  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────╯
➜  ~ coffee remote add folgore-git https://github.com/coffee-tools/folgore.git
✗ Fetch remote from https://github.com/coffee-tools/folgore.git error: Error while add remote: code: 1, msg: repository with name: folgore-git already exists
✗ Error: code: 1, msg: repository with name: folgore-git already exists
```

tareknaser:
> Find a bug when I try to add two times the same repo but in different network

Good point. We need to modify `coffee.add_remote()` so it checks against remotes specific to the network instead of global repositories

vincenzopalazzo:
>Good point. We need to modify coffee.add_remote() so it checks against remotes specific to the network instead of global repositories

Thanks for suggesting the fix, I will try to look at this tomorrow :)

vincenzopalazzo:
MH we should push this forward!

Issue #236, opened 2024-02-07
ci integration testint is broken


vincenzopalazzo:
Due that lightningd/plugin change in the guidance and also the policy our integration testing is broken, we should find a solution for it

cc @tareknaser

tareknaser:
Yes
most of our integration testing depends on `summary` and `helpme` as sample python plugins
we should find another python plugin that is easy to test with cln as well

Maybe [monitor](https://github.com/lightningd/plugins/tree/master/monitor)?

vincenzopalazzo:
I was wondering if we should look inside other directory if a directory is not a plugin directory :/

This make the indexing a little bit more complex but sounds a good solution also for the future?

PR #240, opened 2024-02-10
Add `coffee disable` and `coffee enable` Commands

### LINKED ISSUE #102: Introduce `coffee disable` and `coffee enable` Commands
# Background
Currently, `coffee` provides functionality for managing plugins within the Core Lightning. Users can easily install, configure, and manage plugins using `coffee`'s intuitive commands. 
However, there are scenarios where users may need to disable all plugins installed by `coffee`. This can be necessary for a variety of reasons, such as debugging or troubleshooting issues related to specific plugins or temporarily reverting to a clean Core Lightning environment.

# Goal
The goal of this enhancement is to introduce a new `coffee` command that allows users to disable all plugins previously installed by `coffee`. This feature will provide users with greater flexibility and control over their Core Lightning setup. By running this command, users can effectively "turn off" all plugins, ensuring a clean and predictable Core Lightning environment.

# Example Usage
```shell
# To disable all plugins installed by coffee
coffee disable

# To enable all plugins installed by coffee
coffee enable
```

# Room for Improvement
While the initial implementation focuses on enabling or disabling all plugins at once, there is room for improvement by allowing users to disable individual plugins. To achieve this, the enhancement could introduce the ability to disable a specific plugin using a command like:
```shell
coffee disable <plugin-name>
```

tareknaser:
# Description
This PR adds support for `coffee disable` and `coffee enable` commands, allowing users to toggle plugin status without complete uninstallation. 
The implementation ensures backward compatibility by making the new field "enabled" optional in the plugin struct.

Closes #102
# Changes made
- Implemented `disable` and `enable` methods for `CoffeeManager`
- Added CLI commands `coffee enable/disable <plugin_name>`
- Included HTTP endpoints for the new commands
- Updated documentation
# To Do
- Integration testing for both `coffee` and `coffee_httpd`, to be added following #236
- Add `coffee disable --all` to disable all plugins with a single command

### DIFF
--- a/coffee_cmd/src/cmd.rs
+++ b/coffee_cmd/src/cmd.rs
@@ -73,6 +73,12 @@ pub enum CoffeeCommand {
     /// tipping a plugins developer.
     #[clap(arg_required_else_help = false)]
     Tip { plugin: String, amount_msat: u64 },
+    /// Disable a plugin
+    #[clap(arg_required_else_help = true)]
+    Disable { plugin: String },
+    /// Enable a plugin
+    #[clap(arg_required_else_help = true)]
+    Enable { plugin: String },
 }
 
 #[derive(Debug, Subcommand)]
@@ -114,6 +120,8 @@ impl From<&CoffeeCommand> for coffee_core::CoffeeOperation {
                 plugin,
                 amount_msat,
             } => Self::Tip(plugin.to_owned(), amount_msat.clone()),
+            CoffeeCommand::Disable { plugin } => Self::Disable(plugin.to_owned()),
+            CoffeeCommand::Enable { plugin } => Self::Enable(plugin.to_owned()),
         }
     }
 }
--- a/coffee_cmd/src/coffee_term/command_show.rs
+++ b/coffee_cmd/src/coffee_term/command_show.rs
@@ -14,22 +14,31 @@ pub fn show_list(coffee_list: Result<CoffeeList, CoffeeError>) -> Result<(), Cof
 
     term::println(
         term::format::bold("●"),
-        term::format::tertiary("Plugin installed"),
+        term::format::tertiary("Plugins installed"),
     );
     let mut table = radicle_term::Table::new(TableOptions::bordered());
     table.push([
         term::format::dim(String::from("●")),
         term::format::bold(String::from("Language")),
         term::format::bold(String::from("Name")),
+        term::format::bold(String::from("Enabled?")),
         term::format::bold(String::from("Exec path")),
     ]);
     table.divider();
 
     for plugin in &remotes.plugins {
+        // Get whether the plugin is enabled
+        // If enabled is None, it means the plugin is enabled by default for backward compatibility.
+        let enabled = plugin.enabled.unwrap_or(true);
         table.push([
             term::format::positive("●").into(),
             term::format::highlight(plugin.lang.to_string()),
             term::format::bold(plugin.name()),
+            if enabled {
+                term::format::positive("yes").into()
+            } else {
+                term::format::negative("no").into()
+            },
             term::format::highlight(plugin.exec_path.to_owned()),
         ])
     }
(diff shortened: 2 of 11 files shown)


coffee_cmd/src/coffee_term/command_show.rs
```diff
@@ -14,22 +14,31 @@ pub fn show_list(coffee_list: Result<CoffeeList, CoffeeError>) -> Result<(), Cof
...
         term::format::bold("●"),
-        term::format::tertiary("Plugin installed"),
+        term::format::tertiary("Plugins installed"),
     );
     let mut table = radicle_term::Table::new(TableOptions::bordered());
     table.push([
         term::format::dim(String::from("●")),
         term::format::bold(String::from("Language")),
         term::format::bold(String::from("Name")),
+        term::format::bold(String::from("Enabled?")),
```
vincenzopalazzo, line 24: please remove the `?`

coffee_lib/src/plugin.rs
```diff
@@ -104,6 +104,8 @@ pub struct Plugin {
     conf: Option<Conf>,
     /// FIXME: this field shouldn't be optional
     pub commit: Option<String>,
+    /// Optional for now to be backward compatible
```
vincenzopalazzo, line 107: This is not useful as a doc comment, the doc comment should explain what the field is, and then add more context on why it is optional

coffee_cmd/src/cmd.rs
```diff
@@ -73,6 +73,12 @@ pub enum CoffeeCommand {
     /// tipping a plugins developer.
     #[clap(arg_required_else_help = false)]
     Tip { plugin: String, amount_msat: u64 },
+    /// Disable a plugin
+    #[clap(arg_required_else_help = true)]
+    Disable { plugin: String },
+    /// Enable a plugin
+    #[clap(arg_required_else_help = true)]
+    Enable { plugin: String },
```
vincenzopalazzo, line 81: I was more happy if the enable and disable command was on two different commit

coffee_core/src/coffee.rs
```diff
@@ -597,6 +597,68 @@ impl PluginManager for CoffeeManager {
...
+            .ok_or(error!(
+                "No plugin with name `{plugin}` found in the plugins installed"
+            ))?;
+        log::debug!("plugin: {:?}", plugin);
+        if plugin.enabled == Some(false) {
+            return Err(error!("Plugin `{plugin}` is already disabled"));
+        }
+        self.coffee_cln_config
+            .rm_conf("plugin", Some(&plugin.exec_path))
+            .map_err(|err| error!("{}", err.cause))?;
```
vincenzopalazzo, line 618: nit: cln has an option that it is `disable-plugin=PLUGIN` with the following docs

>If PLUGIN contains a /, plugins with the same path as PLUGIN will
not be loaded at startup. Otherwise, no plugin with that base name will
be loaded at startup, whatever directory it is in. This option is useful for
disabling a single plugin inside a directory. You can still explicitly
load plugins which have been disabled, using lightning-plugin(7) start.

coffee_core/src/coffee.rs
```diff
@@ -597,6 +597,68 @@ impl PluginManager for CoffeeManager {
...
+                "No plugin with name `{plugin}` found in the plugins installed"
+            ))?;
+        log::debug!("plugin: {:?}", plugin);
+        if plugin.enabled.is_none() || plugin.enabled == Some(true) {
+            return Err(error!(
+                "Plugin `{plugin}` is already enabled or enabled by default"
+            ));
+        }
+        self.coffee_cln_config
+            .add_conf("plugin", &plugin.exec_path)
```
vincenzopalazzo, line 649: you here remove the `disable-plugin` that it is much safe

vincenzopalazzo (changes requested):
we are missing test (we can avoid the httpd test btw) for this feature

coffee_core/src/coffee.rs
tareknaser: Yes that's much better
Thanks

coffee_cmd/src/cmd.rs
tareknaser: That would have been better, yes.

vincenzopalazzo (approved):
LGTM, thanks!

PR #242, opened 2024-02-23
Add Branch Selection for `coffee_install()`

### LINKED ISSUE #205: Add Branch Specification for Plugin Installation
# Description:
The objective of this enhancement is to let users specify a branch when installing plugins using `coffee`. 
> It is important to note that this proposal assumes the presence of only one plugin in each remote repository.

# Goal:
- Allow users to install a plugin from a specific branch using the command:
```shell
coffee install <plugin_name> --branch <branch_name>`
```

# Existing Features:
- The `repository` struct already includes a `branch` field.
- A function, `fast_forward(path: &str, branch: &str) -> Result<UpgradeStatus, CoffeeError>`, is available to facilitate fast-forwarding to a specified branch.

tareknaser:
## Description

Add branch selection for `coffee_install` and display plugin git HEAD in `coffee list`.
Fixes #205

## Changes
- Added `Repository::git_head()` to retrieve the repository's git HEAD.
- Added an optional `branch` parameter to `coffee_install()` for branch selection during installation.
- Implemented `git_upgrade_with_git_head()` for checkout using a specific git HEAD.
- Updated `cmd`, `coffee_plugin`, and `coffee httpd` to pass the branch value.
- Modified `coffee list` output to display the git HEAD of plugins.

## Testing
```bash
coffee install <plugin_name> --branch <branch_name>
```

### DIFF
--- a/coffee_cmd/src/coffee_term/command_show.rs
+++ b/coffee_cmd/src/coffee_term/command_show.rs
@@ -22,6 +22,7 @@ pub fn show_list(coffee_list: Result<CoffeeList, CoffeeError>) -> Result<(), Cof
         term::format::bold(String::from("Language")),
         term::format::bold(String::from("Name")),
         term::format::bold(String::from("Enabled")),
+        term::format::bold(String::from("Git HEAD")),
         term::format::bold(String::from("Exec path")),
     ]);
     table.divider();
@@ -30,6 +31,8 @@ pub fn show_list(coffee_list: Result<CoffeeList, CoffeeError>) -> Result<(), Cof
         // Get whether the plugin is enabled
         // If enabled is None, it means the plugin is enabled by default for backward compatibility.
         let enabled = plugin.enabled.unwrap_or(true);
+        let mut commit = plugin.commit.clone().unwrap_or_default();
+        commit = commit.chars().take(7).collect::<String>();
         table.push([
             term::format::positive("●").into(),
             term::format::highlight(plugin.lang.to_string()),
@@ -39,6 +42,7 @@ pub fn show_list(coffee_list: Result<CoffeeList, CoffeeError>) -> Result<(), Cof
             } else {
                 term::format::negative("no").into()
             },
+            term::format::primary(commit),
             term::format::highlight(plugin.exec_path.to_owned()),
         ])
     }
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -15,6 +15,7 @@ use serde_json::json;
 use tokio::process::Command;
 
 use coffee_github::repository::Github;
+use coffee_github::{git_upgrade_with_branch, git_upgrade_with_git_head};
 use coffee_lib::errors::CoffeeError;
 use coffee_lib::plugin_manager::PluginManager;
 use coffee_lib::repository::Repository;
@@ -246,12 +247,36 @@ impl PluginManager for CoffeeManager {
         plugin: &str,
         verbose: bool,
         try_dynamic: bool,
+        branch: Option<String>,
     ) -> Result<(), CoffeeError> {
         log::debug!("installing plugin: {plugin}");
         // keep track if the plugin is successfully installed
         for repo in self.repos.values() {
             if let Some(mut plugin) = repo.get_plugin_by_name(plugin) {
                 log::trace!("{:?}", plugin);
+
+                // if the branch is specified, we need to save current git_head
+                // and checkout the branch
+                let current_git_head = repo.git_head();
+                let current_plugin_commit = plugin.commit.clone();
+                if let Some(branch) = branch.clone() {
+                    log::debug!("checking out branch: {branch}");
+                    if current_git_head.is_none() {
+                        return Err(error!("unable to get the current git head"));
+                    }
+                    let updated_status =
+                        git_upgrade_with_branch(repo.url().path_string.as_str(), &branch, verbose)
+                            .await?;
+                    log::debug!("updated status: {:?}", updated_status);
+
+                    let repository = git2::Repository::open(repo.url().path_string.as_str())
+                        .map_err(|err| error!("{}", err.message()))?;
+
+                    // get commit id after the checkout
+                    let commit = commit_id!(repository);
+                    log::debug!("commit id after the checkout: {commit}");
+                    plugin.commit = Some(commit);
+                }
 
                 // old_root_path is the path where the plugin is cloned and currently stored
                 // eg. ~/.coffee/repositories/<repo_name>/<plugin_name>
@@ -291,7 +316,7 @@ impl PluginManager for CoffeeManager {
                         if !try_dynamic {
                             // mark the plugin enabled
                             plugin.enabled = Some(true);
-                            self.config.plugins.push(plugin);
+                            self.config.plugins.push(plugin.clone());
                             log::debug!("path coffee conf: {}", self.coffee_cln_config.path);
                             self.coffee_cln_config
                                 .add_conf("plugin", &path.to_owned())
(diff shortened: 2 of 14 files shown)


coffee_lib/src/repository.rs
```diff
@@ -40,5 +40,8 @@ pub trait Repository: Any {
     /// return the url of the repository.
     fn url(&self) -> URL;
 
+    /// return the git head of the repository.
+    fn git_head(&self) -> Option<String>;
+
```
vincenzopalazzo, line 45: does not make sense to have this inside the repository trait

coffee_core/src/coffee.rs
```diff
@@ -246,13 +247,37 @@ impl PluginManager for CoffeeManager {
...
+                    log::debug!("updated status: {:?}", updated_status);
+
+                    let repository = git2::Repository::open(repo.url().path_string.as_str())
+                        .map_err(|err| error!("{}", err.message()))?;
+
+                    // get commit id after the checkout
+                    let commit = commit_id!(repository);
+                    log::debug!("commit id after the checkout: {commit}");
+                    plugin.commit = Some(commit);
+                }
```
vincenzopalazzo, line 279: IMHO to much code for just `sh!("git checkout {branch}")`?

coffee_core/src/coffee.rs
```diff
@@ -302,6 +327,23 @@ impl PluginManager for CoffeeManager {
                         } else {
                             self.start_plugin(&path).await?;
                         }
+
+                        if branch.is_some() {
+                            // we can safely unwrap here because we have checked it before
+                            let current_git_head = current_git_head.unwrap();
```
vincenzopalazzo, line 333: the safety comment start with `// SAFETY: ...`

coffee_core/src/coffee.rs
```diff
@@ -302,6 +327,23 @@ impl PluginManager for CoffeeManager {
...
+                            let updated_status = git_upgrade_with_git_head(
+                                repo.url().path_string.as_str(),
+                                &current_git_head,
+                                verbose,
+                            )
+                            .await?;
+                            log::debug!("updated status: {:?}", updated_status);
+
+                            plugin.commit = current_plugin_commit;
+                        }
```
vincenzopalazzo, line 345: wow this install method is starting to be 2 pages long?

coffee_github/src/utils.rs
```diff
@@ -25,7 +25,7 @@ pub async fn clone_recursive_fix(repo: git2::Repository, url: &URL) -> Result<()
     Ok(())
 }
 
-pub async fn git_upgrade(
+pub async fn git_upgrade_with_branch(
```
vincenzopalazzo, line 28: why changing the name here?

coffee_github/src/utils.rs
```diff
@@ -48,3 +48,27 @@ pub async fn git_upgrade(
...
+    git_head: &str,
+    verbose: bool,
+) -> Result<UpgradeStatus, CoffeeError> {
+    use tokio::process::Command;
+
+    let repo = git2::Repository::open(path).map_err(|err| error!("{}", err.message()))?;
+
+    let (local_commit, _) = get_repo_info!(repo);
+
+    let mut cmd = format!("git fetch origin\n");
```
vincenzopalazzo, line 63: `let mut cmd = "git fetch origin\n";`

`format!` is not needed

coffee_github/src/utils.rs
```diff
@@ -48,3 +48,27 @@ pub async fn git_upgrade(
         Ok(UpgradeStatus::Updated(upstream_commit, date))
     }
 }
+
+pub async fn git_upgrade_with_git_head(
+    path: &str,
+    git_head: &str,
+    verbose: bool,
+) -> Result<UpgradeStatus, CoffeeError> {
+    use tokio::process::Command;
```
vincenzopalazzo, line 57: idk if this is a good idea, the checkout is just `git checkout <branch-name>` 

Why we need to use all this code?

coffee_cmd/src/coffee_term/command_show.rs
```diff
@@ -22,6 +22,7 @@ pub fn show_list(coffee_list: Result<CoffeeList, CoffeeError>) -> Result<(), Cof
         term::format::bold(String::from("Language")),
         term::format::bold(String::from("Name")),
         term::format::bold(String::from("Enabled")),
+        term::format::bold(String::from("Git HEAD")),
```
vincenzopalazzo, line 25: a user need also a git PHD to understand what is a git HEAD? Idk if this is necessary.

docs/docs-book/src/using-coffee.md
```diff
@@ -104,6 +104,14 @@ To install a plugin statically, you simply need to run.
...
 ```
 
+#### Specifying a branch for installation
+
+To install a plugin from a specific branch, you simply need to run.
+
+```bash
+coffee install <plugin_name> --branch <branch_name>
+```
+
```
vincenzopalazzo, line 114: we should not document every single command possibility otherwise our documentation will be always uncompleted

vincenzopalazzo (changes requested):
I think that this implementation is a little bit too complex for what we need to do

coffee_lib/src/repository.rs
tareknaser: good point. I will update it

coffee_core/src/coffee.rs
tareknaser: I can see some potential issues with simply using `git checkout {branch}`:
- If the branch is new (not fetched yet), we need to `git fetch origin` first. 
	- To make sure we are returning to the expected state after switching back, we need the repository's Git HEAD, not just the current branch. This is because the branch might have been updated remotely (when we ran `git fetch origin`), and we want to go back to the exact commit the user was on.
	-  We need to verify that the repository's Git HEAD information exists (introduced in a previous PR as an Option to be backward compatible) 
- The plugin's `commit` field is updated during installation and then reverted afterward. Otherwise, we have false information.

Do you think that any of these steps is redundant?

coffee_core/src/coffee.rs
tareknaser: I agree!
We can keep `coffee install`  for just installing and configuring plugins and introduce `coffee upgrade --plugin <plugin_name> --branch <branch_name>` for upgrades and branch switching

coffee_github/src/utils.rs
tareknaser: because I added another function that uses the git commit to upgrade

coffee_cmd/src/coffee_term/command_show.rs
tareknaser: We already print the git HEAD to the user with `coffee remote list`
If we add the possibility to manage the branch for the plugin in separation from the repository, I think we have to print it in `coffee list`

docs/docs-book/src/using-coffee.md
tareknaser: Right. I can remove this section

coffee_core/src/coffee.rs
vincenzopalazzo: No, my point is that the code that we are writing the code of the install is not good at all. Please see some of my PR that I am refactoring it.

In addition, both install and upgrade command will run the same logic and with your code we should duplicate the checkout logic. This is error prone

coffee_core/src/coffee.rs
tareknaser: `git_upgrade_with_branch()` does `git fetch origin`

coffee_github/src/utils.rs
vincenzopalazzo: I do not see the point of changing the name, due that there is no git upgrade command in `git`

vincenzopalazzo: Moreover, git uses always a branch there is never a `without_branch` cases

docs/docs-book/src/using-coffee.md
vincenzopalazzo: but we should came up with a way to document optional parameters

coffee_cmd/src/coffee_term/command_show.rs
tareknaser: Sure. Do you want me to rename "Git HEAD" to "commit" in `coffee remote list` output too?

vincenzopalazzo:
Closing in favor of https://github.com/coffee-tools/coffee/pull/248

PR #243, opened 2024-02-25
install command: add -T option to cp to not create nested folders on duplicate calls

daywalker90:
If you call coffee install more than once it will create folders like this:

1. ~/.coffee/<network>/plugins/<plugin_name>
2. ~/.coffee/<network>/plugins/<plugin_name>/<plugin_name>

the -T option for cp should fix this.

### DIFF
--- a/coffee_core/src/coffee.rs
+++ b/coffee_core/src/coffee.rs
@@ -270,7 +270,7 @@ impl PluginManager for CoffeeManager {
                     old_root_path,
                     new_root_path
                 );
-                let script = format!("cp -r {old_root_path} {new_root_path}");
+                let script = format!("cp -r -T {old_root_path} {new_root_path}");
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
LGTM waiting for @tareknaser that is the author of this feature for confimation

tareknaser (approved):
Thanks for the catch!

vincenzopalazzo:
>Just noticed that -T is a GNU cp feature and won't work on macOS.

you had an OSX iirc? did you not test on your machine before approving the PR?

tareknaser:
> you had an OSX iirc? did you not test on your machine before approving the PR?

No I actually missed this one. I read the documentation for the flag online and thought it made perfect sense for a quick fix 
My bad

PR #244, opened 2024-02-25
Always print install errors

daywalker90:
When running `coffee install` and there is an error e.g. the plugin could not be started by cln, coffee does not print an error. Not even in `verbose` mode. Only with the `RUST_LOG` env you can eventually see what happens. I propose to always show an install error to the user who just follows the documentation.

### DIFF
--- a/coffee_cmd/src/main.rs
+++ b/coffee_cmd/src/main.rs
@@ -26,15 +26,15 @@ async fn run(args: CoffeeArgs, mut coffee: CoffeeManager) -> Result<(), CoffeeEr
             } else {
                 None
             };
-            let result = coffee.install(&plugin, verbose, dynamic).await;
-            if let Some(spinner) = spinner {
-                if result.is_ok() {
-                    spinner.finish();
-                } else {
-                    spinner.failed();
+            match coffee.install(&plugin, verbose, dynamic).await {
+                Ok(_) => {
+                    spinner.and_then(|spinner| Some(spinner.finish()));
+                    term::success!("Plugin {plugin} Compiled and Installed")
                 }
-            } else if result.is_ok() {
-                term::success!("Plugin {plugin} Compiled and Installed")
+                Err(err) => {
+                    spinner.and_then(|spinner| Some(spinner.failed()));
+                    term::error(format!("{err}"))
+                }
             }
         }
         CoffeeCommand::Remove { plugin } => {


coffee_cmd/src/main.rs
```diff
@@ -27,14 +27,20 @@ async fn run(args: CoffeeArgs, mut coffee: CoffeeManager) -> Result<(), CoffeeEr
                 None
             };
             let result = coffee.install(&plugin, verbose, dynamic).await;
```
vincenzopalazzo, line 29:
```suggestion
            let result = coffee.install(&plugin, verbose, dynamic).await?;
```

I think just this is enough? to get the error back?

daywalker90: Well yes, but then the code after needs changes anyways:
```
coffee.install(&plugin, verbose, dynamic).await?;
if let Some(spinner) = spinner {
    spinner.finish();
} else {
    term::success!("Plugin {plugin} Compiled and Installed")
}
```
I can force push this version if you like it more.

vincenzopalazzo: yes I like your diff except if we can do `spinner.and_then(|spinner| spinnet.finish());`

daywalker90: which diff do you like 1. or 2. ? Also no clue why and where you want that line. It would need to be 
```rust
spinner.and_then(|spin| Some(spin.finish()));
```
and even then clippy is yelling at me.

vincenzopalazzo: >and even then clippy is yelling at me.

No I do not think so we do something similar inside the upgrade command, IMHO the install code should be the same.

This is a diff that works

```diff 
diff --git a/coffee_cmd/src/main.rs b/coffee_cmd/src/main.rs
index d5708fa..238e897 100644
--- a/coffee_cmd/src/main.rs
+++ b/coffee_cmd/src/main.rs
@@ -26,15 +26,15 @@ async fn run(args: CoffeeArgs, mut coffee: CoffeeManager) -> Result<(), CoffeeEr
             } else {
                 None
             };
-            let result = coffee.install(&plugin, verbose, dynamic).await;
-            if let Some(spinner) = spinner {
-                if result.is_ok() {
-                    spinner.finish();
-                } else {
-                    spinner.failed();
+            match coffee.install(&plugin, verbose, dynamic).await {
+                Ok(_) => {
+                    spinner.and_then(|splinner| Some(splinner.finish()));
+                    term::success!("Plugin {plugin} Compiled and Installed")
+                }
+                Err(err) => {
+                    spinner.and_then(|spinner| Some(spinner.failed()));
+                    return Err(err);
                 }
-            } else if result.is_ok() {
-                term::success!("Plugin {plugin} Compiled and Installed")
             }
         }
         CoffeeCommand::Remove { plugin } => {
```

Sorry about this but I do not like too much branching in Rust for readability reasons

If clippy has something to say, we silence it :) and we make it as a repository rule; Everyone need to have a convention stile agreement

daywalker90: Ah got it. One thing tho this code ignores the verbose variable and also does a `term::success` even if `verbose=false´:
```rust
            match coffee.install(&plugin, verbose, dynamic).await {
                Ok(_) => {
                    spinner.and_then(|spinner| Some(spinner.failed()));
                    if verbose {
                        term::success!("Plugin {plugin} Compiled and Installed")
                    }
                }
                Err(err) => {
                    spinner.and_then(|spinner| Some(spinner.failed()));
                    term::error(format!("{err}"))
                }
            }
```

coffee_cmd/src/main.rs
```diff
@@ -26,15 +26,17 @@ async fn run(args: CoffeeArgs, mut coffee: CoffeeManager) -> Result<(), CoffeeEr
...
-                if result.is_ok() {
-                    spinner.finish();
-                } else {
-                    spinner.failed();
+            match coffee.install(&plugin, verbose, dynamic).await {
+                Ok(_) => {
+                    spinner.and_then(|spinner| Some(spinner.finish()));
+                    if verbose {
+                        term::success!("Plugin {plugin} Compiled and Installed")
+                    }
```
vincenzopalazzo, line 34:
```suggestion
```

nit: With the verbose flag we skip just the spinner, because does not make sense inside the terminal state.

vincenzopalazzo (changes requested):

coffee_cmd/src/main.rs
daywalker90: Then you get no notification of a success in verbose mode ?!

vincenzopalazzo: >Then you get no notification of a success in verbose mode ?!

The opposite? you get a notification in both cases?

daywalker90: I don't get it. Your "Suggested Change" is removing the lines that inform the user in verbose mode. Then the only thing left that indicates success or error is the spinner. But the spinner is only active in non-verbose mode.

daywalker90: Ok now it makes the confusion a little clearer (your change has a `-` infront of all 3 lines!). One thing left: my code is not a behavioural change for the `term::success`. Before it was only run in verbose mode aswell. This is because it was only run if `spinner` was `None` and `spinner` was only `None` if `verbose==true`:
```
let spinner = if !verbose {
    Some(...)
} else {
    None
}
if let Some(spinner) = spinner {
    ...
}  else if result.is_ok() {
    term::success!("Plugin {plugin} Compiled and Installed")
}

vincenzopalazzo: I told you branching is shit with rust.

This was a bug, so please remove the if and print `term::success!("Plugin {plugin} Compiled and Installed")` always

vincenzopalazzo (approved):
Thanks for catching this out!

PR #248, opened 2024-02-26
git: adding basic function to checkout a branch

vincenzopalazzo:
Alternative version of https://github.com/coffee-tools/coffee/pull/242

### DIFF
--- a/coffee_github/src/utils.rs
+++ b/coffee_github/src/utils.rs
@@ -1,8 +1,10 @@
+use log::debug;
+use tokio::process::Command;
+
 use coffee_lib::errors::CoffeeError;
 use coffee_lib::macros::error;
 use coffee_lib::url::URL;
 use coffee_lib::{commit_id, get_repo_info, sh};
-use log::debug;
 
 use coffee_lib::types::response::UpgradeStatus;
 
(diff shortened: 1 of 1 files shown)


coffee_github/src/utils.rs
```diff
@@ -1,8 +1,10 @@
+use log::debug;
+use tokio::process::Command;
```
vincenzopalazzo, line 2: wrong import place

coffee_core/src/coffee.rs
```diff
@@ -285,6 +287,13 @@ impl PluginManager for CoffeeManager {
                         plugin.root_path = new_root_path;
                         plugin.exec_path = new_exec_path;
 
+                        if let Some(branch) = branch {
+                            // FIXME: Where we store the date? how we manage it?
+                            let (commit, _) =
+                                git_checkout(&plugin.root_path, &branch, verbose).await?;
```
tareknaser, line 293: We switch the whole repository to the required branch but we don't switch back to the original branch of the repository after installing the plugin

tareknaser (changes requested):
Thanks

Here are some notes as well:
- We need to update integration tests in `tests/src/coffee_integration_tests.rs` to pass `branch` parameter to `coffee_install`
- coffee plugin needs to pass the `branch` value as well. See `coffee_plugin/src/plugin/plugin_mod.rs` 
- We can also add a comment in the docs to mention the new parameter

coffee_core/src/coffee.rs
vincenzopalazzo: nice point

vincenzopalazzo:
Ok, this needs to go in before the first release, to fix the CI.

We can add the remote with a specific commit when summary was sane

tareknaser:
> We can add the remote with a specific commit when summary was sane

Is this possible with git?
if not, we can force push it manually in the tests

What do you think?

vincenzopalazzo:
>Is this possible with git?
if not, we can force push it manually in the tests

Yeah make sense to do it

PR #249, opened 2024-02-26
Checkt That Network Specific Plugin Path Doesn't Exist

tareknaser:
This PR checks that the destination path doesn't exist when we copy the plugin folder from the repositories folder to the destination network specific folder. If it exists, we remove it


vincenzopalazzo:
Please see https://github.com/coffee-tools/coffee/pull/250

This patch is an hack not a real solution

Issue #245, opened 2024-02-26
Better support for dynamic and important plugins


daywalker90:
I identified a few issues related to the title:

- dynamic plugins are not added to the internal state, so things like `list`or `remove` don't work
- help text should make clear that `enable` and `disable` are for static plugins only and make a  `start` and `stop` command for dynamic plugins 
- - **OR** overload `enable` and `disable` for dynamic plugins to be their `start` and `stop` 
- using `remove` should stop a dynamic plugin
- plugin authors need to be able to mark their plugin `important` in the manifest and coffee should support that

I can make pr's for these if you like, but i wanted to gather some feedback first.

vincenzopalazzo:
There are really good points thanks, I think we should implement these too before our 0.1.0-rc1 because looks like basic functionalities

PR #251, opened 2024-02-27
feat(core): consistent dynamic plugins install/remove

daywalker90:
- dynamic plugins now get saved persistently
- on remove dynamic plugins get stopped
- do not return on every Err, if it's expected

part of #245


vincenzopalazzo (approved):
LGTM thanks

PR #252, opened 2024-02-27
feat: support important plugins

daywalker90:
added a new manifest field to mark plugin as important and only install it as static and important

part of #245

### DIFF
--- a/coffee_lib/src/plugin.rs
+++ b/coffee_lib/src/plugin.rs
@@ -165,6 +165,18 @@ impl Plugin {
     pub fn tipping_info(&self) -> Option<Tipping> {
         self.conf.as_ref().and_then(|conf| conf.tipping.clone())
     }
+
+    pub fn important(&self) -> bool {
+        if let Some(config) = &self.conf {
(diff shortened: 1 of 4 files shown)


coffee_lib/src/plugin.rs
```diff
@@ -165,6 +165,19 @@ impl Plugin {
...
+        let config = if let Some(conf) = &self.conf {
+            conf
+        } else {
+            return false;
+        };
+        if let Some(important) = config.plugin.important {
+            important
+        } else {
+            false
+        }
```
vincenzopalazzo, line 179:
```suggestion
        let config = if let Some(conf) = &self.conf {
            if let Some(important) = config.plugin.important {
                important
             } else {
                 false
             }
        } else {
            return false;
        };
```

We can still do better but at least this is cleaner

vincenzopalazzo (changes requested):
Just one commit and we are ready to go, thanks

vincenzopalazzo (approved):
LGTM, thanks!

PR #253, opened 2024-02-27
fix: tip function sent unwanted amount and expected wrong response

daywalker90:
When fetching an invoice with an amount and then paying that invoice cln will error if you've sent an amount on the pay command aswell.

Tip also expected an incompatible `CoffeeTip` response from `pay`. `pay` responses don't have a field `for_plugin` or `invoice`

tested with cln 24.02rc2


vincenzopalazzo (approved):
LGTM thanks for testing and patching it

PR #265, opened 2024-04-12
core: remove repo path from url

vincenzopalazzo:
This is a preparation work to simplify the PR https://github.com/coffee-tools/coffee/pull/239

looking right now if it is possible to remove the plugin path from the struct, but looks like a little bit tricky? maybe no


vincenzopalazzo:
Mh idk if this works!

PR #266, opened 2024-04-12
core: fix the repo name from url

vincenzopalazzo:
A bug that I found while working on the coffee refactoring


coffee_lib/src/url.rs
```diff
@@ -18,7 +18,7 @@ pub struct URL {
 }
 
 /// Handle GitHub HTTP links
-fn remove_dot_git_from_url(url: &str) -> &str {
+pub(crate) fn remove_dot_git_from_url(url: &str) -> &str {
```
tareknaser, line 21: Why do we need to add `pub(crate)`?
I don't see it used anywhere other than this file

tareknaser (approved):
Thanks

coffee_lib/src/url.rs
vincenzopalazzo: it is used inside the tests, and the test is another module

PR #267, opened 2024-04-14
Rename Command to Get Available Plugins in a Remote

### LINKED ISSUE #261: remote: add a new command to inspect the plugins inside the remote
Currently, it is really hard to see the plugins inside a remote, so I will propose a new command called `inspect` that works like `coffee remote --plugins folgore-git list`

So currently to see the plugins inside a directory you need to do

```
➜  ~ coffee remote --plugins folgore-git list
● Plugins installed
```

but IMHO this should be a little bit better, an example:

```
➜  ~ coffee remote inspect folgore-git
● Plugins installed
```

`inspect` is more intuitive

What do you think @tareknaser

tareknaser:
This commit changes the CLI command to get available plugins in a remote repository in coffee. 

Previously the user had to run `coffee remote <repository_name> --plugins` which was confusing. This commit changes it to be `coffee remote inspect <repository_name>` and updates documentation.

Fixes #261

### DIFF
--- a/coffee_cmd/src/cmd.rs
+++ b/coffee_cmd/src/cmd.rs
@@ -48,8 +48,6 @@ pub enum CoffeeCommand {
     Remote {
         #[clap(subcommand)]
         action: Option<RemoteAction>,
-        #[arg(short, long, action = clap::ArgAction::SetTrue, requires = "remote-name")]
-        plugins: bool,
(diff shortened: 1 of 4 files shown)


vincenzopalazzo (approved):
LGTM thanks

